PR #2, opened 2020-11-22
Adding skeleton code for the daemon and a testing client

### LINKED ISSUE #1: Update build to include pre-alpha daemon

tjayrush:
Closes #1.

This is ready to merge. I left the `kv.proto` work as it did earlier, but for the non-kv protos, they are built into the build folder.


canepat:
@tjayrush I've tried to build this branch locally but fails with the same error reported by CI job `linux-clang-coverage`:
```
CMake Error at silkrpc/CMakeLists.txt:94 (file):
  file COPY cannot find "/home/builder/build/eth.pb.cc": No such file or
  directory.
```

canepat:
Moreover I would like to discuss where we should place things in the repo (maybe we could write a guideline after we reach an agreement) because IMHO `notes` folder should be placed outside `silkrpc` subtree.

tjayrush:
> @tjayrush I've tried to build this branch locally but fails with the same error reported by CI job `linux-clang-coverage`:
> 
> ```
> CMake Error at silkrpc/CMakeLists.txt:94 (file):
>   file COPY cannot find "/home/builder/build/eth.pb.cc": No such file or
>   directory.
> ```
I think I should use a `custom_command` here. `file(COPY` (I guess) is a `cmake` time command (it runs during the `cmake`). After `make clean`, the files don't exists and they do not exist until `protoc` creates them (which happens during `make`), so of course `cmake` can't find them.

I commented this out for now. We can discuss tomorrow.

tjayrush:
> Moreover I would like to discuss where we should place things in the repo (maybe we could write a guideline after we reach an agreement) because IMHO `notes` folder should be placed outside `silkrpc` subtree.

I'm totally open to re-arranging the files in any way that works. In fact, `notes` is really only for my use, so maybe we can put it in the `.gitignore` or something.

Again, we can discuss tomorrow.

PR #9, opened 2020-11-25
Add skeleton code for RPC server

### LINKED ISSUE #1: Update build to include pre-alpha daemon

tjayrush:
Closes #1 

This new PR is identical to the previous PR.


.circleci/config.yml
```diff
@@ -40,8 +40,14 @@ commands:
           key: *hunter-cache-key
           paths:
             - ~/.hunter
+      # Build the silkrpc folder first so the generated files are available for the other folders
       - run:
           name: "Build"
+          working_directory: ~/build
+          command: cmake --build ~/build/silkrpc
```
canepat, line 47: Is this necessary for the generation of additional `proto` binding files? If so, please see my comments below

cmd/CMakeLists.txt
```diff
@@ -42,6 +43,17 @@ add_executable(kv_seek kv_seek.cpp ${SILKRPC_SRC})
 target_include_directories(kv_seek PRIVATE ${CMAKE_SOURCE_DIR})
 target_link_libraries(kv_seek absl::flags_parse gRPC::grpc++_unsecure protobuf::libprotobuf silkworm)
 
+# test only client
+file(GLOB_RECURSE GENERATED_SOURCES CONFIGURE_DEPENDS "${CMAKE_BINARY_DIR}/*.cc")
+message(STATUS "CMAKE_BINARY_DIR: " ${CMAKE_SOURCE_DIR})
+message(STATUS "CMAKE_BINARY_DIR: " ${CMAKE_BINARY_DIR})
+message(STATUS "GENERATED_SOURCES: " ${SILKRPC_SOURCES})
+message(STATUS "GENERATED_SOURCES: " ${GENERATED_SOURCES})
```
canepat, line 51: We could remove the `message` logs (maybe added for debug) in order to maintain the build output as clean as possibile

cmd/test_client.cpp
```diff
@@ -0,0 +1,101 @@
...
+int main(int argc, char** argv) {
+    auto credentials = grpc::InsecureChannelCredentials();
+    auto channel = grpc::CreateChannel("localhost:50051", credentials);
+
+    EthServiceClient EthService(channel);
+    std::cout << "\n\t" << EthService.GetBlock(12) << std::endl;
+
+    return 0;
+}
+#endif
```
canepat, line 101: If I understand correctly and this `test_client` is testing code for the `eth.proto` interface, please see below my comment for the `eth.proto` file.

silkrpc/main.cpp
```diff
@@ -14,6 +14,207 @@
...
+    "\"sha3Uncles\": \"0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347\",\n"
+    "\"size\": \"0x219\",\n"
+    "\"stateRoot\": \"0x821c41f30a2fd9580605363784a8a2a6575b255ec37cacf87fe52715b8828d8e\",\n"
+    "\"timestamp\": \"0x55ba42c0\",\n"
+    "\"totalDifficulty\": \"0x33f2ffe033\",\n"
+    "\"transactions\": [],\n"
+    "\"transactionsRoot\": \"0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421\",\n"
+    "\"uncles\": []}";
+
+#endif
```
canepat, line 220: I wouldn't put this server implementation related to the `eth.proto` interface in the `Silkrpc` main, please see below my comment for the `eth.proto` file.

silkrpc/protos/eth.proto
```diff
@@ -0,0 +1,47 @@
...
+
+// Structure used to handle requests for eth_getTrans* related calls
+message GetTransactionRequest {
+  uint64 block = 1;
+  uint64 tx_id = 3;
+  string hash = 5;
+}
+
+// The returned data from eth_getTrans* related requests
+message GetTransactionResponse { string message = 1; }
```
canepat, line 47: The goals of [Silkrpc](https://github.com/torquem-ch/silkrpc/blob/master/docs/ARCHITECTURE.md) project are _both_ 1) producing a C++ implementation of the existing `ETH JSON RPC` interface IF#2 separated from but connected to Turbo-Geth/Silkworm Core via IF#1 _and_ 2) exploring the possibility to define an improved extended interface IF#3, probably binary for increased performance.

This `eth.proto` file and the related build files plus code (test_client, main) belong to the goal 2), which is still under discussion in terms of design, i.e. what IF#3 interface should offer exactly and how. 

So IMHO the content of this PR could stay in a dedicated branch until we finish the exploration of 2) and the design is finished with related decisions taken (e.g. which primitives are exposed through IF#3? do we need two different daemons?), just in order to maintain the `master` branch as clean as possibile.

PR #42, opened 2021-04-26
Continuous integration using Github Actions

canepat:


canepat:
Eventually we will merge Silkrpc project into Silkworm, where https://github.com/torquem-ch/silkworm/pull/834 already introduced support for Windows and macOS CI jobs using GitHub Actions.

PR #45, opened 2021-04-28
Update silkworm submodule to 5122e8a

canepat:


canepat:
Silkworm does not build with GCC 10.2 at commit 5122e8a

PR #50, opened 2021-05-04
Eth get uncle by block number and index

Sixtysixter:
Implementation of the API eth_getUncleByBlockNumberAndIndex

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -22,7 +22,7 @@ The following table shows the current [JSON RPC API](https://eth.wiki/json-rpc/A
 | eth_getBlockTransactionCountByHash      | Yes          |                                            |
 | eth_getBlockTransactionCountByNumber    | Yes          |                                            |
 | eth_getUncleByBlockHashAndIndex         | -            | not yet implemented                        |
-| eth_getUncleByBlockNumberAndIndex       | -            | not yet implemented                        |
+| eth_getUncleByBlockNumberAndIndex       | Yes          |                                            |
 | eth_getUncleCountByBlockHash            | -            | not yet implemented                        |
 | eth_getUncleCountByBlockNumber          | -            | not yet implemented                        |
 |                                         |              |                                            |
--- a/silkrpc/commands/eth_api.cpp
+++ b/silkrpc/commands/eth_api.cpp
@@ -296,12 +296,38 @@ asio::awaitable<void> EthereumRpcApi::handle_eth_get_uncle_by_block_hash_and_ind
 
 // https://eth.wiki/json-rpc/API#eth_getunclebyblocknumberandindex
 asio::awaitable<void> EthereumRpcApi::handle_eth_get_uncle_by_block_number_and_index(const nlohmann::json& request, nlohmann::json& reply) {
+    auto params = request["params"];
+    if (params.size() != 2) {
+        auto error_msg = "invalid eth_getUncleByBlockNumberAndIndex params: " + params.dump();
+        SILKRPC_ERROR << error_msg << "\n";
+        reply = make_json_error(request["id"], 100, error_msg);
+        co_return;
+    }
+    const auto block_id = params[0].get<std::string>();
+    const auto index = params[1].get<std::string>();
+    SILKRPC_DEBUG << "block_id: " << block_id << " index: " << index << "\n";
+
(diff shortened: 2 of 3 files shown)


silkrpc/commands/eth_api.cpp
```diff
@@ -296,12 +296,40 @@ asio::awaitable<void> EthereumRpcApi::handle_eth_get_uncle_by_block_hash_and_ind
...
+        auto idx = std::stoul(index);
+        if (idx >= ommers.size()) {
+            reply = make_json_content(request["id"], nullptr);
+            co_return;
+        }
+        const auto total_difficulty = co_await core::rawdb::read_total_difficulty(tx_database, block_with_hash.hash, block_number);
+        auto uncle = ommers[idx];
+
+        const silkworm::BlockBody block_body{};
+        const evmc::bytes32 hash {};
```
canepat, line 328: Lines 327 and 328 are unnecessary, see comment below

silkrpc/commands/eth_api.cpp
```diff
@@ -296,12 +296,40 @@ asio::awaitable<void> EthereumRpcApi::handle_eth_get_uncle_by_block_hash_and_ind
...
+        silkworm::BlockWithHash block_hash{{block_body, uncle}, hash};
```
canepat, line 329: Some improvements here:
1) naming seems ambiguous (it's not an hash) and too similar with line 316
2) it should be more clear that we're creating an empty block to hold the uncle block header
3) `hash` is empty here, whilst it should be `uncle.hash()`

In short I suggest to replace line 329 with:

`silkworm::BlockWithHash uncle_block_with_hash{{{}, uncle}, uncle.hash()};`

silkrpc/commands/eth_api.cpp
```diff
@@ -296,12 +296,40 @@ asio::awaitable<void> EthereumRpcApi::handle_eth_get_uncle_by_block_hash_and_ind
...
+        const Block new_block{block_hash, total_difficulty};
```
canepat, line 330: `new_block` -> `uncle_block_with_hash_and_td`

(`silkrpc::Block` should be renamed something like `silkrpc::BlockWithHashAndTD` but let's do this in a separate PR)

silkrpc/commands/eth_api.cpp
```diff
@@ -296,12 +296,40 @@ asio::awaitable<void> EthereumRpcApi::handle_eth_get_uncle_by_block_hash_and_ind
...
 
     try {
         ethdb::TransactionDatabase tx_database{*tx};
 
-        reply = make_json_content(request["id"],  "0x" + to_hex_no_leading_zeros(0));
+        const auto block_number = co_await core::get_block_number(block_id, tx_database);
+        const auto block_with_hash = co_await core::rawdb::read_block_by_number(tx_database, block_number);
+        const auto ommers = block_with_hash.block.ommers;
+
+        auto idx = std::stoul(index);
```
canepat, line 319: Have you tried with `index` different from `0x0` (e.g. `0x1`)? This line doesn't work, it should be:

`auto idx = std::stoul(index, 0, 16);`

because `index` is in QUANTITY format (we should introduce a new `silkrpc::Quantity` type but we'll do in a separate PR)

canepat (changes requested):

docs/API.md
```diff
@@ -28,7 +28,7 @@ The following table shows the current [JSON RPC API](https://eth.wiki/json-rpc/A
 |                                         |              |                                            |
 | eth_getTransactionByHash                | -            | not yet implemented                        |
 | eth_getTransactionByBlockHashAndIndex   | -            | not yet implemented                        |
-| eth_getTransactionByBlockNumberAndIndex | -            | not yet implemented                        |
+| eth_getTransactionByBlockNumberAndIndex | Yes          |                                            |
```
canepat, line 31: Please add in this PR also `eth_getUncleCountByBlockNumber` implementation because it's closely related to this one

Sixtysixter: oopps.

silkrpc/commands/eth_api.cpp
Sixtysixter: Done

canepat (approved):

PR #61, opened 2021-05-08
eth get transaction by block hash and index

Sixtysixter:
Implementation of the API eth_getTransactionByBlockHashAndIndex


silkrpc/commands/eth_api.cpp
```diff
@@ -471,8 +496,7 @@ asio::awaitable<void> EthereumRpcApi::handle_eth_get_transaction_by_block_hash_a
     }
 
     co_await tx->close(); // RAII not (yet) available with coroutines
-    co_return;
-}
+    co_return;}
```
canepat, line 499: Curly brace should be on new line as it was previously

canepat (changes requested):

canepat (approved):

PR #80, opened 2021-06-02
eth get transaction by hash

Sixtysixter:
Implementation of eth_getTransactionByHash API method


docs/API.md
```diff
@@ -27,7 +27,7 @@ The following table shows the current [JSON RPC API](https://eth.wiki/json-rpc/A
 | eth_getUncleCountByBlockNumber          | Yes          |                                            |
 | eth_getBlockReceipts                    | Yes          | same as parity_getBlockReceipts            |
 |                                         |              |                                            |
-| eth_getTransactionByHash                | -            | not yet implemented                        |
+| eth_getTransactionByHash                | Partially    | Finalized transaction not supported        |
```
canepat, line 30: You can put `Yes` as value for `Availability` and `partially implemented` in `Notes`

silkrpc/commands/eth_api.cpp
```diff
@@ -447,17 +447,34 @@ asio::awaitable<void> EthereumRpcApi::handle_eth_get_uncle_count_by_block_number
...
+            // TODO
+            // Maybe no finalized transaction, try to retrieve it from the pool
+            SILKRPC_DEBUG << "Retrieving not finalized transactions from pool not implemented yet\n";
+            reply = make_json_content(request["id"], nullptr);
+        } else {
+            reply = make_json_content(request["id"], *optional_transaction);
+        }
     } catch (const std::exception& e) {
-        SILKRPC_ERROR << "exception: " << e.what() << " processing request: " << request.dump() << "\n";
+        SILKRPC_ERROR << "exception: " << e.what() << "\n";
```
canepat, line 474: I would leave the line as in master here:
```
SILKRPC_ERROR << "exception: " << e.what() << " processing request: " << request.dump() << "\n";
```
because it produces information about the request causing the failure, it helps reproducing the issue

silkrpc/commands/eth_api.cpp
```diff
@@ -447,17 +447,34 @@ asio::awaitable<void> EthereumRpcApi::handle_eth_get_uncle_count_by_block_number
...
         reply = make_json_error(request["id"], 100, e.what());
     } catch (...) {
-        SILKRPC_ERROR << "unexpected exception processing request: " << request.dump() << "\n";
+        SILKRPC_ERROR << "unexpected exception\n";
```
canepat, line 477: Same as comment above

silkrpc/core/rawdb/chain.cpp
```diff
@@ -202,6 +202,31 @@ asio::awaitable<silkworm::Bytes> read_body_rlp(const DatabaseReader& reader, con
     co_return data;
 }
 
+asio::awaitable<std::optional<silkrpc::Transaction>> read_transaction_by_hash(const DatabaseReader& reader, const evmc::bytes32& transaction_hash) {
+    const silkworm::ByteView tx_hash{transaction_hash.bytes, silkworm::kHashLength};
+
+    auto bytes = co_await reader.get_one(silkworm::db::table::kTxLookup.name, tx_hash);
+    if (bytes.length() == 0) {
```
canepat, line 209: Probably `bytes.empty()` is more readable

silkrpc/core/rawdb/chain.cpp
```diff
@@ -202,6 +202,31 @@ asio::awaitable<silkworm::Bytes> read_body_rlp(const DatabaseReader& reader, con
...
+    }
+    auto block_number = std::stoul(silkworm::to_hex(bytes), 0, 16);
+    SILKRPC_TRACE << "Block number " << block_number << " for transaction hash " << transaction_hash << "\n";
+    const auto block_with_hash = co_await core::rawdb::read_block_by_number(reader, block_number);
+    const auto transactions = block_with_hash.block.transactions;
+
+    for (size_t idx{0}; idx < transactions.size(); idx++) {
+        auto ethash_hash{hash_of_transaction(transactions[idx])};
+        silkworm::ByteView hash_view{ethash_hash.bytes, silkworm::kHashLength};
+        SILKRPC_TRACE << "tx " << idx << ") hash: " << silkworm::to_bytes32(hash_view) << "]\n";
```
canepat, line 221: Missing `(` before `idx` value

silkrpc/commands/eth_api.cpp
```diff
@@ -447,17 +447,34 @@ asio::awaitable<void> EthereumRpcApi::handle_eth_get_uncle_count_by_block_number
...
+        const auto optional_transaction = co_await core::rawdb::read_transaction_by_hash(tx_database, transaction_hash);
+        if (!optional_transaction) {
+            // TODO
+            // Maybe no finalized transaction, try to retrieve it from the pool
+            SILKRPC_DEBUG << "Retrieving not finalized transactions from pool not implemented yet\n";
+            reply = make_json_content(request["id"], nullptr);
+        } else {
+            reply = make_json_content(request["id"], *optional_transaction);
+        }
     } catch (const std::exception& e) {
```
canepat, line 473: First of all exception `std::invalid_argument` should be catched to handle possible failures coming from internal functions, e.g. see [line 104 in chain.cpp](https://github.com/torquem-ch/silkrpc/blob/master/silkrpc/core/rawdb/chain.cpp#L104).

This mechanism (I mean throwing `std::invalid_argument` instead of using `std::optional` for data not found) is far from ideal and will be subject to change in the future, but it allows for clean implementation of higher-level methods (that otherwise would suffer from repeated resource-closing code blocks in absence of RAII support in coroutines).

Please check carefully the return value spec in this case, probably empty value (not error)

canepat (changes requested):

silkrpc/commands/eth_api.cpp
Sixtysixter: The `std::optional` management is required as the `core::rawdb::read_transaction_by_hash` should return a null `Transaction` as it could be in not yer finalized block.  For that reason the call does't throw a `std::invalid_argument` exception if a `Transaction` is not found for the given hash.
I could add a catch for `std::invalid_argument` for any future modification (and I did).

canepat (approved):

PR #82, opened 2021-06-03
Silkrpc-as-a-library example

canepat:
Add Silkrpc-as-a-library example [WIP]
Improve CppLint coverage


tjayrush:
I'm pretty excited to try this out some day. If only I could find the time.

PR #95, opened 2021-06-16
eth get transaction receipt

Sixtysixter:
Implementation of eth_getTransactionReceipt API method


silkrpc/json/types.cpp
```diff
@@ -305,20 +306,36 @@ void to_json(nlohmann::json& json, const Receipt& receipt) {
...
-            throw std::system_error{std::make_error_code(std::errc::invalid_argument), "Receipt CBOR: number expected in [2]"};
+        receipt.success = json[start_index] == 1u;
+
+        start_index++;
+        if (!json[start_index].is_number()) {
+            auto message = str(boost::format("Receipt CBOR: number expected in [%d]") % start_index);
+            throw std::system_error{std::make_error_code(std::errc::invalid_argument), message};
         }
-        receipt.cumulative_gas_used = json[2];
+        receipt.cumulative_gas_used = json[start_index];
```
canepat, line 338: I would go with the hard-coded solution handling 4 CBOR fields because 1) we really don't need backward compatibility with old format 2) readability is worse this way IMHO 3) we avoid introducing dependency on boost::format

canepat (changes requested):

silkrpc/commands/eth_api.cpp
```diff
@@ -581,12 +581,46 @@ asio::awaitable<void> EthereumRpcApi::handle_eth_get_transaction_by_block_number
...
+        reply = make_json_content(request["id"], nullptr);
+        const auto block_with_hash = co_await core::rawdb::read_block_by_transaction_hash(tx_database, transaction_hash);
+        auto receipts = co_await core::get_receipts(tx_database, block_with_hash.hash, block_with_hash.block.header.number);
 
-        reply = make_json_content(request["id"], to_quantity(0));
+        auto transactions = block_with_hash.block.transactions;
+        if (receipts.size() != transactions.size()) {
+            throw std::invalid_argument{"Unexpected size for receipts in handle_eth_get_transaction_receipt"};
+        }
+        // const silkworm::ByteView tx_hash{transaction_hash.bytes, silkworm::kHashLength};
```
canepat, line 605: Remove comment

silkrpc/commands/eth_api.cpp
```diff
@@ -581,12 +581,46 @@ asio::awaitable<void> EthereumRpcApi::handle_eth_get_transaction_by_block_number
...
+        size_t tx_index = -1;
+        for (size_t idx{0}; idx < transactions.size(); idx++) {
+            auto ethash_hash{hash_of_transaction(transactions[idx])};
+            // silkworm::ByteView hash_view{ethash_hash.bytes, silkworm::kHashLength};
```
canepat, line 609: Remove comment

silkrpc/commands/eth_api.cpp
```diff
@@ -581,12 +581,46 @@ asio::awaitable<void> EthereumRpcApi::handle_eth_get_transaction_by_block_number
...
+            SILKRPC_TRACE << "tx " << idx << ") hash: " << silkworm::to_bytes32(ethash_hash.bytes) << "\n";
+            if (memcmp(transaction_hash.bytes, ethash_hash.bytes, silkworm::kHashLength) == 0) {
```
canepat, line 611: I suggested to use `operator==` for `evmc::bytes32` type here but your solution is _smarter_ because it avoids one useless copy! Just a couple of tiny requested changes: 1) add <cstring> inclusion 2) use `std::memcmp` instead of `memcmp`

canepat (changes requested):

canepat (approved):
LGTM

PR #132, opened 2021-07-30
Eth call fix

lupin012:
Fix per eth_call
1) Modified 2 errors code
2) Add check to return insuf funds
3) Add handler in json format to avoid exception in case invalid UTF8


silkrpc/core/evm_executor.cpp
```diff
@@ -168,34 +168,51 @@ asio::awaitable<ExecutionResult> EVMExecutor::call(const silkworm::Block& block,
...
+
+                if (have < want) {
+                   silkworm::Bytes data{};
+                   std::string s1 = "insufficient funds for gas * price + value: address 0x";
+                   std::string addrValue = silkworm::to_hex(*txn.from);
+                   std::string haveStr = " have ";
+                   std::string haveValue = intx::to_string(have);
+                   std::string wantStr = " want ";
+                   std::string wantValue = intx::to_string(want);
+                   std::string error_message = s1 + addrValue + haveStr + haveValue + wantStr + wantValue;
```
canepat, line 182: I would replace lines 176-182 with:
```
std::string from = silkworm::to_hex(*txn.from);
std::string error = "insufficient funds for gas * price + value: address 0x" + from + " have " + intx::to_string(have) + " want " + intx::to_string(want);
```
to improve overall readability of method: the line becomes quite long but I would keep the error message formatting as compact as possible because it is not so relevant

silkrpc/core/evm_executor.hpp
```diff
@@ -39,6 +39,7 @@ struct ExecutionResult {
     int64_t error_code;
     uint64_t gas_left;
     silkworm::Bytes data;
+    std::optional<std::string> preCheckErrorMessage;
```
canepat, line 42: `preCheckErrorMessage` -> `pre_check_error{std::nullopt}`

Two little things here:
- name should be changed for naming conformance with Silkworm and the rest of Silkrpc
- `std::optional` should be initialized in conformance with Silkworm and the rest of Silkrpc

canepat (changes requested):

silkrpc/commands/eth_api.cpp
```diff
@@ -836,6 +836,8 @@ asio::awaitable<void> EthereumRpcApi::handle_eth_call(const nlohmann::json& requ
 
         if (execution_result.error_code == evmc_status_code::EVMC_SUCCESS) {
             reply = make_json_content(request["id"], "0x" + silkworm::to_hex(execution_result.data));
+        } else if (execution_result.preCheckErrorMessage) {
+                reply = make_json_error(request["id"], -32000, execution_result.preCheckErrorMessage.value());
```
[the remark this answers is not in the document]
lupin012: done

silkrpc/core/evm_executor.cpp
lupin012: done

silkrpc/core/evm_executor.hpp
lupin012: done

lupin012: done

PR #137, opened 2021-08-10
Eth call fix2

lupin012:
Modified json interface to be compliant with EIP : Eip2930, Eip1559


silkrpc/json/types.cpp
```diff
@@ -150,22 +153,44 @@ void to_json(nlohmann::json& json, const Transaction& transaction) {
...
     json["s"] = silkrpc::to_quantity(silkworm::rlp::big_endian(transaction.s));
 }
 
+
+
 } // namespace silkworm
 
 namespace silkrpc {
 
+uint64_t get_block_sizes(const Block &b) {
```
canepat, line 179: I would rename it `get_block_size` and put it in `Block` struct

canepat (changes requested):
Add missing unit tests to avoid coverage diff

silkrpc/json/types.cpp
```diff
@@ -150,22 +153,44 @@ void to_json(nlohmann::json& json, const Transaction& transaction) {
         json["to"] =  nullptr;
     }
     json["type"] = silkrpc::to_quantity(transaction.type.value_or(0));
+
+    if (transaction.type && transaction.type == silkworm::kEip1559TransactionType) {
```
canepat, line 157: First leg of AND is redundant

lupin012: ok

silkrpc/json/types.cpp
lupin012: renamed, but Block is struct of silkworm not silkrpc.

canepat: Don't think so as it's defined in `silkrpc` namespace...

lupin012: Sorry my mistake

canepat (approved):

PR #140, opened 2021-08-20
Eth call fix3

lupin012:
This PR contains:
- fix on insufficient balance computation
- fix on default gas value if not present
- refactory evm_executor with add pre_check method
- aligns integration tests results


silkrpc/types/call.hpp
```diff
@@ -42,7 +42,7 @@ struct Call {
         silkworm::Transaction txn{};
         txn.from = from;
         txn.to = to;
-        txn.gas_limit = gas.value_or(0);
+        txn.gas_limit = gas.value_or(30000000);
```
canepat, line 45: Add missing unit test for this as suggested by Codecov

silkrpc/core/evm_executor.cpp
```diff
@@ -167,27 +192,15 @@ asio::awaitable<ExecutionResult> EVMExecutor::call(const silkworm::Block& block,
...
-                   asio::post(*context_.io_context, [exec_result, self = std::move(self)]() mutable {
-                      self.complete(exec_result);
-                   });
-                   return;
-                }
+                const intx::uint128 g0{silkworm::intrinsic_gas(txn, rev >= EVMC_HOMESTEAD, rev >= EVMC_ISTANBUL)};
+                assert(g0 <= UINT64_MAX); // true due to the precondition (transaction must be valid)
 
-                if (have < want) {
+                const auto error = pre_check(rev, txn, base_fee_per_gas, state, want, g0);
```
canepat, line 200: Passing `silkworm::EVM` as parameter to `pre_check` allows removing both `rev` and `state` that can be obtained from `evm` as follows:
```
const evmc_revision rev{evm.revision()};
const silkworm::IntraBlockState state{evm.state()};
```

canepat (changes requested):
Empty line 224 in `evm_executor.cpp` should be removed
Line 230 in `evm_executor.cpp` has bad indentation

silkrpc/core/evm_executor.cpp
lupin012: ok

silkrpc/core/evm_executor.cpp
```diff
@@ -152,6 +152,31 @@ std::string EVMExecutor::get_error_message(int64_t error_code, const silkworm::B
     return error_message;
 }
 
+std::optional<std::string> pre_check(const evmc_revision rev, const silkworm::Transaction& txn, const intx::uint256 base_fee_per_gas, const silkworm::IntraBlockState & state,
```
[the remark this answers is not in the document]
lupin012: > ok

silkrpc/types/call.hpp
lupin012: few tests added

canepat (approved):

PR #150, opened 2021-08-30
Eth call fix4

lupin012:
- modified checks in pre_check()
- avoid to call effective_gas_price() if max_fee_per_gas  AND max_priority_fee_per_gas  are zero


canepat (approved):
Just a couple of things: 1) generally we should try to avoid Codecov remarks as much as possible 2) we should use Catch2 `TEST/SECTION` feature also to avoid code duplication among tests (`evm_executor_test.cpp` contains a lot of duplicated code that can be factorized at `SECTION` level)

PR #157, opened 2021-09-09
BlockNumberOrHash implementation

Sixtysixter:
The new class BlockNumberOrHash has been added.


silkrpc/core/rawdb/chain.cpp
```diff
@@ -98,6 +98,19 @@ asio::awaitable<intx::uint256> read_total_difficulty(const DatabaseReader& reade
...
+    if (bnoh.is_number()) {
+        co_return co_await read_block_by_number(reader, bnoh.number());
+    } else if (bnoh.is_hash()) {
+        co_return co_await read_block_by_hash(reader, bnoh.hash());
+    } else if (bnoh.is_tag()) {
+        auto block_number = co_await get_latest_block_number(reader);
+        co_return co_await read_block_by_number(reader, block_number);
+    }
+
+    throw std::runtime_error{"invalid block_number_or_hash value"};
```
canepat, line 111: Not sure if this case should be handled as exception or by return value containing `std::optional`. In the end of the day it all depends on how this function is supposed to be used and what must be done when `bnoh` is invalid: I guess that an error should be returned, so probably it's fine as it is

silkrpc/types/block.hpp
```diff
@@ -35,6 +37,70 @@ struct Block : public silkworm::BlockWithHash {
...
+        return is_tag() ? *tag_ : "";
+    }
+
+private:
+    void build(std::string const& bnoh);
+    void set_number(std::string const& input, int base);
+
+    std::unique_ptr<std::uint64_t> number_;
+    std::unique_ptr<evmc::bytes32> hash_;
+    std::unique_ptr<std::string> tag_;
```
canepat, line 99: I would really consider the usage of `std::variant` here: whenever possible, we should try to use types already provided in the standard library

silkrpc/types/block.hpp
```diff
@@ -35,6 +37,70 @@ struct Block : public silkworm::BlockWithHash {
 
 std::ostream& operator<<(std::ostream& out, const Block& b);
 
+class BlockNumberOrHash {
+public:
+    BlockNumberOrHash() {}
```
canepat, line 42: You could remove default constructor and undefined state, that in my understanding are present just for deserialization from `nlohmann::json` objects, by applying a specific technique outlined [here](https://github.com/nlohmann/json#how-can-i-use-get-for-non-default-constructiblenon-copyable-types)

canepat (changes requested):

silkrpc/core/rawdb/chain.cpp
Sixtysixter: This solution has been chosen as it seems in line with other `read_*` functions. that return an exception if something went wrong. 
Returning an `std::optional` introduce a little difference...

silkrpc/types/block.hpp
Sixtysixter: knowledge is power!! 
Done.

silkrpc/types/block.hpp
Sixtysixter: Done

canepat (approved):

PR #216, opened 2022-01-12
support clang versions >= 10.0.0 and compiler version check

### LINKED ISSUE #11: Build files intermittently
11:21 AM] tjayrush: Do you ever see this build failure?

[ 76%] Linking CXX executable kv_seek_async
Undefined symbols for architecture x86_64:
  "___llvm_profile_runtime", referenced from:
      ___llvm_profile_runtime_user in libsilkwormd.a(util.cpp.o)
     (maybe you meant: ___llvm_profile_runtime_user)
[11:22 AM] tjayrush: The only way I can make it disappear is to do rm -fR * in the build folder and then cmake .. and make. That cleans it up, but then it keeps coming back.
[11:23 AM] canepat: Unfortunately no
[11:24 AM] canepat: I have no clue what is means atm
[11:24 AM] canepat: Seems an issue in Silkworm linking
[11:25 AM] canepat: Have you tried googling the error?
[11:26 AM] canepat: Seems that Silkworm references ___llvm_profile_runtime instead of ___llvm_profile_runtime_user
[11:26 AM] tjayrush: I have another weird error with the linking (weak symbol)... I googled, but not must useful.
[11:27 AM] tjayrush: It's weird because it's intermittent and disappears if I remove the build artifacts.
[11:27 AM] canepat: You are on a Mac, aren't you?
[11:27 AM] tjayrush: I'll play around with it. I thought you may have seen it.
[11:27 AM] tjayrush: Yes. Mac. You're linux?
[11:28 AM] canepat: Yes Ubuntu 20.04
[11:29 AM] canepat: https://github.com/Swinject/Swinject/issues/296#issuecomment-357711213
GitHub
Undefined symbols for architecture x86_64: ___llvm_profile_runtime ...
I&#39;m getting an issue linking Swinject framework into my project when using Xcode9 and Swift3.2 It&#39;s giving me the following error: Undefined symbols for architecture x86_64: &qu...
[11:31 AM] canepat: Could be something around the -fprofile-instr-generate setting, but it seems Siilkworm does it right by passing it both as compiler and as linker option
[11:32 AM] canepat: But above all: it's not the first time you build, you already built many times, so something changed
[11:33 AM] canepat: What did you have changed recently? (time locality principle also for bugs... :grin:)
[11:35 AM] tjayrush: I do make clean all the time...silkworm gets re-built each time I do.
[11:36 AM] tjayrush: "time locality principle also for bugs..." Yes.
[11:37 AM] tjayrush: Let me look into it. Thanks for the pointers... I think I'll write an issue just so I don't forget.

Giulio2002:
Fixes #11


canepat:
I've moved the automatic compiler check after CMake project and added support for both GCC and Clang

@Giulio2002 can you please try if this still works on Mac Clang after the changes?

canepat (approved):

PR #224, opened 2022-01-25
Fix intermittent unit test fail for Clang release

### LINKED ISSUE #186: Intermittent unit test failure using Clang in release
```
~/workspace/silkrpc$ ./build_clang_release.sh 
[  1%] Built target hex
[  1%] Built target instructions
[  3%] Built target cborcpp
[  4%] Built target generate_sentry_grpc
[  4%] Built target secp256k1
[  5%] Built target mdbx-static
[  7%] Built target loader
[ 14%] Built target ff
[ 16%] Built target mdbx_copy
[ 17%] Built target mdbx_chk
[ 17%] Built target tooling
[ 20%] Built target roaring
[ 23%] Built target evmone
[ 23%] Built target mdbx_drop
[ 24%] Built target mdbx_dump
[ 28%] Built target silkinterfaces
[ 28%] Built target mdbx_load
[ 29%] Built target mdbx_stat
[ 42%] Built target silkworm_core
[ 57%] Built target silkworm_node
[ 72%] Built target silkrpc
[ 72%] Built target kv_seek_async
[ 74%] Built target kv_seek
[ 75%] Built target ethbackend_async
[ 75%] Built target kv_seek_async_callback
[ 76%] Built target kv_seek_async_asio
[ 77%] Built target kv_seek_async_coroutines
[ 78%] Built target silkrpcdaemon
[ 78%] Built target kv_seek_both
[ 78%] Built target ethbackend
[ 78%] Built target ethbackend_coroutines
[ 78%] Built target get_latest_block
[100%] Built target unit_test
tcmalloc: large alloc 139668265140224 bytes == (nil) @  0x7f070d783680 ...
terminating with uncaught exception of type std::bad_alloc: std::bad_alloc

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
unit_test is a Catch v2.13.4 host application.
Run with -? for options

-------------------------------------------------------------------------------
RemoteCursor::seek
  success w/ sync read - sync write
-------------------------------------------------------------------------------
~/workspace/silkrpc/silkrpc/ethdb/kv/remote_cursor_test.cpp:362
...............................................................................

~/workspace/silkrpc/silkrpc/ethdb/kv/remote_cursor_test.cpp:362: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal

===============================================================================
test cases:  82 |  81 passed | 1 failed
assertions: 678 | 677 passed | 1 failed

./build_clang_release.sh: line 6: 465118 Aborted                 (core dumped) cmd/unit_test
```

The unit test failure is intermittent (i.e typically once in a series of 50 unit_test executions), it happens only using Clang compiler in release (no issue using Clang debug and GCC release/debug) and can be associated to different test cases or sections.

The problem can be always tracked down to the usage of some coroutine API expecting one or more `const silkworm::ByteView&` parameter activated using `asio::co_spawn`. The solution here is modifying all pass-by-reference usages of `silkworm::ByteView` into pass-by-value (extra verification required).

canepat:
Remove pass-by-ref for silkworm::ByteView
Improve unit test loop runner
Fixes #186

### DIFF
--- a/silkrpc/commands/debug_api_test.cpp
+++ b/silkrpc/commands/debug_api_test.cpp
@@ -60,7 +60,7 @@ public:
         co_return;
     }
 
-    asio::awaitable<KeyValue> seek(const silkworm::ByteView& key) override {
+    asio::awaitable<KeyValue> seek(silkworm::ByteView key) override {
         const auto key_ = silkworm::to_hex(key);
 
         KeyValue out;
@@ -81,7 +81,7 @@ public:
         co_return out;
     }
 
-    asio::awaitable<KeyValue> seek_exact(const silkworm::ByteView& key) override {
+    asio::awaitable<KeyValue> seek_exact(silkworm::ByteView key) override {
         const nlohmann::json table = json_.value(table_name_, empty);
         const auto& entry = table.value(silkworm::to_hex(key), "");
         auto value{*silkworm::from_hex(entry)};
@@ -103,7 +103,7 @@ public:
         co_return out;
     }
 
-    asio::awaitable<silkworm::Bytes> seek_both(const silkworm::ByteView& key, const silkworm::ByteView& value) override {
+    asio::awaitable<silkworm::Bytes> seek_both(silkworm::ByteView key, silkworm::ByteView value) override {
         silkworm::Bytes key_{key};
         key_ += value;
 
@@ -114,7 +114,7 @@ public:
         co_return out;
     }
 
-    asio::awaitable<KeyValue> seek_both_exact(const silkworm::ByteView& key, const silkworm::ByteView& value) override {
+    asio::awaitable<KeyValue> seek_both_exact(silkworm::ByteView key, silkworm::ByteView value) override {
         silkworm::Bytes key_{key};
         key_ += value;
 
--- a/silkrpc/core/blocks_test.cpp
+++ b/silkrpc/core/blocks_test.cpp
@@ -25,6 +25,7 @@
 #include <gmock/gmock.h>
 #include <silkworm/common/base.hpp>
 
+#include <silkrpc/common/log.hpp>
 #include <silkrpc/stagedsync/stages.hpp>
 #include <silkrpc/ethdb/tables.hpp>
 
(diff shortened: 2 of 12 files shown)


silkrpc/commands/debug_api_test.cpp
```diff
@@ -60,7 +60,7 @@ class DummyCursor : public silkrpc::ethdb::CursorDupSort {
         co_return;
     }
 
-    asio::awaitable<KeyValue> seek(const silkworm::ByteView& key) override {
+    asio::awaitable<KeyValue> seek(silkworm::ByteView key) override {
```
Giulio2002, line 63: why does passing by reference cause unit tests to fail?

silkrpc/core/blocks_test.cpp
```diff
@@ -44,38 +45,44 @@ class MockDatabaseReader : public core::rawdb::DatabaseReader {
...
 
 TEST_CASE("get_block_number", "[silkrpc][core][blocks]") {
+    SILKRPC_LOG_STREAMS(null_stream(), null_stream());
     const silkworm::ByteView kExecutionStage{stages::kExecution};
     MockDatabaseReader db_reader;
     asio::thread_pool pool{1};
 
     SECTION("kEarliestBlockId") {
-        auto result = asio::co_spawn(pool, get_block_number(kEarliestBlockId, db_reader), asio::use_future);
+        static const std::string EARLIEST_BLOCK_ID = kEarliestBlockId;
```
Giulio2002, line 54: how does this change anything, and why must it be static?  it is an unit test, wouldn't this keep the variable in memory indefinetely?

silkrpc/core/blocks_test.cpp
```diff
@@ -44,38 +45,44 @@ class MockDatabaseReader : public core::rawdb::DatabaseReader {
...
     SECTION("number in hex") {
-        auto result = asio::co_spawn(pool, get_block_number("0x12345", db_reader), asio::use_future);
+        static const std::string BLOCK_ID_HEX = "0x12345";
+        auto result = asio::co_spawn(pool, get_block_number(BLOCK_ID_HEX, db_reader), asio::use_future);
         CHECK(result.get() == 0x12345);
     }
 
     SECTION("number in dec") {
-        auto result = asio::co_spawn(pool, get_block_number("67890", db_reader), asio::use_future);
+        static const std::string BLOCK_ID_DEC = "67890";
```
Giulio2002, line 84: why static? and why not directly pass it to the method?

silkrpc/core/blocks_test.cpp
```diff
@@ -44,38 +45,44 @@ class MockDatabaseReader : public core::rawdb::DatabaseReader {
...
             []() -> asio::awaitable<KeyValue> { co_return KeyValue{silkworm::Bytes{}, *silkworm::from_hex("1234567890123456")}; }
         ));
-        auto result = asio::co_spawn(pool, get_block_number(kPendingBlockId, db_reader), asio::use_future);
+        auto result = asio::co_spawn(pool, get_block_number(PENDING_BLOCK_ID, db_reader), asio::use_future);
         CHECK(result.get() == 0x1234567890123456);
     }
 
     SECTION("number in hex") {
-        auto result = asio::co_spawn(pool, get_block_number("0x12345", db_reader), asio::use_future);
+        static const std::string BLOCK_ID_HEX = "0x12345";
```
Giulio2002, line 78: why static? and why not directly pass it to the method?

silkrpc/core/blocks_test.cpp
```diff
@@ -44,38 +45,44 @@ class MockDatabaseReader : public core::rawdb::DatabaseReader {
...
         EXPECT_CALL(db_reader, get(db::table::kSyncStageProgress, kExecutionStage)).WillOnce(InvokeWithoutArgs(
             []() -> asio::awaitable<KeyValue> { co_return KeyValue{silkworm::Bytes{}, *silkworm::from_hex("1234567890123456")}; }
         ));
-        auto result = asio::co_spawn(pool, get_block_number(kLatestBlockId, db_reader), asio::use_future);
+        auto result = asio::co_spawn(pool, get_block_number(LATEST_BLOCK_ID, db_reader), asio::use_future);
         CHECK(result.get() == 0x1234567890123456);
     }
 
     SECTION("kPendingBlockId") {
+        static const std::string PENDING_BLOCK_ID = kPendingBlockId;
```
Giulio2002, line 69: why static? and why not directly pass it to the method?

silkrpc/core/storage_walker_test.cpp
```diff
@@ -63,7 +63,7 @@ class DummyCursor : public silkrpc::ethdb::CursorDupSort {
         co_return;
     }
 
-    asio::awaitable<KeyValue> seek(const silkworm::ByteView& key) override {
+    asio::awaitable<KeyValue> seek(silkworm::ByteView key) override {
```
Giulio2002, line 66: why is passing by reference removed?

silkrpc/ethdb/kv/remote_cursor.cpp
```diff
@@ -31,7 +31,7 @@ asio::awaitable<void> RemoteCursor::open_cursor(const std::string& table_name) {
     co_return;
 }
 
-asio::awaitable<KeyValue> RemoteCursor::seek(const silkworm::ByteView& key) {
+asio::awaitable<KeyValue> RemoteCursor::seek(silkworm::ByteView key) {
```
Giulio2002, line 34: why is passing by reference removed?

silkrpc/commands/debug_api_test.cpp
canepat: because pass-by-ref leads to undefined behavior if the object passed by-ref goes out of scope, so it's a powerful optimization technique in some cases (large objects) but requires extreme caution, see the other conversations for additional information

BTW, `silkworm::ByteView` is a template specialization of `std::basic_string_view` which should be passed by-value idiomatically, see [here](https://isocpp.org/blog/2021/11/three-reasons-to-pass-stdstring-view-by-value-arthur-odwyer?s=09)

silkrpc/core/blocks_test.cpp
canepat: First of all, `get_block_number` is a coroutine taking a `const std::string&` as first formal argument. The calling code in this `SECTION` spawns the coroutine call using `asio::co_spawn` and passing `kEarliestBlockId` (declared `const char*`) as first actual argument. Some automatic conversions and temporary object creations happen here, namely:

1) a temporary `std::string` is created and initialized with a _copy_ of `kEarliestBlockId` content
2) a temporary `std::string_view` representing the entire content of such temporary string is created via conversion operator
3) such temporary string view is passed by constant reference to `get_block_number`

The call to `get_block_number` coroutine will be activated in the `pool` separate thread surely _after_ the end of the `asio::co_spawn` call, which is however the relevant function call to determine the lifetime of the string temporary object created at step 1: hence, such temporary string can be destroyed after the `asio::co_spawn` call and the view can be pointing to nowhere when the `get_block_number` coroutine gets activated. BTW this behavior is undefined and compiler are allowed to perform different kinds of optimizations, in fact the undefined behavior popped just in Clang release mode, GCC was fine

As per `static` specifier, you're right: it's not necessary, I will push a commit to remove it

silkrpc/core/blocks_test.cpp
canepat: Same as above

silkrpc/core/blocks_test.cpp
canepat: Same as above

silkrpc/core/storage_walker_test.cpp
canepat: because it is causing issue #186 with temporary actual argument `silkworm::Bytes{}` lifetime when this coroutine is activated by means of `asio::co_spawn`

silkrpc/ethdb/kv/remote_cursor.cpp
canepat: same as above

Giulio2002 (approved):

PR #226, opened 2022-01-26
JSON benchmark

canepat:


canepat:
The technical path explored in this PR was valuable in terms of performance gains (not game-changing, though) but at the cost of cluttering the code base a lot.

Hence, I'm closing this for now. We can reopen it in the future if/when we will need to focus on improving JSON encoding/decoding performance.

PR #229, opened 2022-02-01
Update version control to 4.1.0

Giulio2002:


canepat:
Two more test cases require changes to enforce version condition limits, so they are basically buggy and should be improved:
- **"KV protocol version major mismatch"**: major number should be 3 instead of 2 at [line 154](https://github.com/torquem-ch/silkrpc/blob/kv-4.1.0/silkrpc/protocol/version_test.cpp#L154) just to test for the nearest lower major version that must fail. The test case can be improved by explicitly testing `KV_SERVICE_API_VERSION.major - 1` and `KV_SERVICE_API_VERSION.major + 1`. This is true also for similar tests for the other interfaces
- **"KV protocol version minor mismatch"**: major number should be 4 instead of 3 at [line 172](https://github.com/torquem-ch/silkrpc/blob/kv-4.1.0/silkrpc/protocol/version_test.cpp#L172). This test case is buggy: it should fail for minor number incompatibility, whilst it seems to me that's failing for major number one

canepat (changes requested):
Two more test cases require changes to enforce version condition limits, so they are basically buggy and should be improved:
- **"KV protocol version major mismatch"**: major number should be 3 instead of 2 at [line 154](https://github.com/torquem-ch/silkrpc/blob/kv-4.1.0/silkrpc/protocol/version_test.cpp#L154) just to test for the nearest lower major version that must fail. The test case can be improved by explicitly testing `KV_SERVICE_API_VERSION.major - 1` and `KV_SERVICE_API_VERSION.major + 1`. This is true also for similar tests for the other interfaces
- **"KV protocol version minor mismatch"**: major number should be 4 instead of 3 at [line 172](https://github.com/torquem-ch/silkrpc/blob/kv-4.1.0/silkrpc/protocol/version_test.cpp#L172). This test case is buggy: it should fail for minor number incompatibility, whilst it seems to me that's failing for major number one

canepat (approved):

PR #237, opened 2022-03-05
Updating silkworm in silkrpc

enriavil1:


canepat (changes requested):
Both clang and gcc CI jobs are broken because Silkworm changed the location of submodules: they have been moved to the `third_party` subfolder, so at least one include in Silkrpc CMakeLists.txt file needs to be updated (see error in CI jobs).

Issue #242, opened 2022-03-12
Rename RemoteBuffer as RemoteState


canepat:
`RemoteState` seems a better name than `RemoteBuffer` for the remote KV-based implementation of the `State` abstraction

enriavil1:
#244

PR #244, opened 2022-03-12
Fixed issue #242

umutondersu:
Changed RemoteBuffer to RemoteState on all paths


silkrpc/core/evm_executor.hpp
```diff
@@ -65,7 +65,7 @@ class EVMExecutor {
     const core::rawdb::DatabaseReader& db_reader_;
     const silkworm::ChainConfig& config_;
     asio::thread_pool& workers_;
-    state::RemoteBuffer buffer_;
+    state::RemoteState buffer_;
```
canepat, line 68: After renaming the variable type `RemoteBuffer` as `RemoteState`, also the variable name `buffer_` should be renamed `remote_state_`

silkrpc/core/remote_buffer.hpp
```diff
@@ -124,10 +124,10 @@ class RemoteBuffer : public silkworm::State {
 
 private:
     asio::io_context& io_context_;
-    AsyncRemoteBuffer async_buffer_;
+    AsyncRemoteState async_buffer_;
```
canepat, line 127: Same as above: probably `async_state_` is now better than `async_buffer_`

silkrpc/core/remote_buffer_test.cpp
```diff
@@ -89,7 +89,7 @@ TEST_CASE("async remote buffer", "[silkrpc][core][remote_buffer]") {
         asio::io_context io_context;
         MockDatabaseReader db_reader;
         const uint64_t block_number = 1'000'000;
-        AsyncRemoteBuffer arb{io_context, db_reader, block_number};
+        AsyncRemoteState arb{io_context, db_reader, block_number};
```
canepat, line 92: Same as above: probably `ars` or `state` could be better now than `arb`

silkrpc/core/remote_buffer_test.cpp
```diff
@@ -116,7 +116,7 @@ TEST_CASE("async remote buffer", "[silkrpc][core][remote_buffer]") {
         MockDatabaseReader db_reader{code};
         const uint64_t block_number = 1'000'000;
         const auto code_hash{0x04491edcd115127caedbd478e2e7895ed80c7847e903431f94f9cfa579cad47f_bytes32};
-        RemoteBuffer remoteBuffer(io_context, db_reader, block_number);
+        RemoteState remoteBuffer(io_context, db_reader, block_number);
```
canepat, line 119: Same as above: probably `remote_state` or `state` could be better now than `remoteBuffer`

canepat (changes requested):
Thanks a lot for contributing! Given that we're changing a couple of type names in this PR, I've just suggested the same changes also in variable names,

enriavil1:
#242

canepat (approved):

canepat:
> #242

Thanks a lot @enriavil1 for bringing this PR to completion!

PR #255, opened 2022-03-21
debug_traceBlockBy[Number|Hash]

Sixtysixter:
Implementation for debug_traceBlockByNumber and debug_traceBlockByHash


canepat (approved):

silkrpc/core/evm_tracing.cpp
```diff
@@ -79,6 +79,15 @@ void to_json(nlohmann::json& json, const Trace& trace) {
...
 }
 
+void to_json(nlohmann::json& json, const std::vector<Trace>& traces) {
+    json = nlohmann::json::array();
+    for (auto& trace : traces) {
+        nlohmann::json entry;
+        to_json(entry, trace);
+        json.push_back(entry);
+    }
+}
```
canepat, line 89: It should be sufficient to define serialization for `Trace` here (i.e. `void to_json(nlohmann::json& json, const Trace& trace)`) because Nlohmann JSON library knows how to handle sequences of serializable types.

PR #269, opened 2022-04-23
Refactor context pool to use one thread per context

canepat:


lupin012 (commented):
Add new ABSL_FLAG  to configure externally WaitMode

lupin012 (changes requested):
Add new ABSL_FLAG to configure waitMode

lupin012 (changes requested):
get_context() : Could be useful understand if instead of round robin policy could be useful insert some policy during delivery as: context with less queued request

silkrpc/context_pool.hpp
```diff
@@ -31,37 +31,253 @@
...
+
+    void join();
 
     void stop();
 
+    void run();
+
     Context& get_context();
 
     asio::io_context& get_io_context();
```
lupin012, line 283: Could be useful leave only one method public (get_io_context or get_context) not both sometimes is called get_context() other get_io_context()

silkrpc/context_pool.hpp
```diff
@@ -31,37 +31,253 @@
...
-    void run();
+    void start();
+
+    void join();
 
     void stop();
 
+    void run();
+
     Context& get_context();
```
lupin012, line 281: instead of get_context() could be useful rename something that explain that we are selecting a context (select_context()) with this name seems only a get method

silkrpc/context_pool.cpp
```diff
@@ -23,83 +23,152 @@
...
-    SILKRPC_DEBUG << "ContextPool::stop completed\n";
+    SILKRPC_TRACE << "ContextPool::stop completed\n";
+}
+
+void ContextPool::run() {
+    start();
+    join();
 }
 
 Context& ContextPool::get_context() {
```
lupin012, line 174: could be useful inserts some policy to select context with less queued request or ...

silkrpc/main.cpp
```diff
@@ -167,6 +167,8 @@ int main(int argc, char* argv[]) {
         };
 
         // Check protocol version compatibility with Core Services
+        SILKRPC_LOG << "Checking protocol version compatibility with core services...\n";
+
```
lupin012, line 171: in context_pool pass wait_mode selected by user

lupin012 (changes requested):
http/server.cpp constructor  could be useful add also socket_.set_option(reuse_port(true));

silkrpc/context_pool.cpp
canepat: I undestand your suggestion, nevertheless I would not add any of them in this PR. The reasons are:
- it is tempting to add smart policies, but simple ones (e.g. counting enqueued requests in each scheduler) are not proven to be correct for real workloads (i.e. incoming requests are different, hence heterogeneous in terms of performance impact)
- some experimentation and measurements for different policies should be done first
- round-robin is a very simple yet effective scheduling policy w/o knowledge of the input distribution

I would prefer for you to open an issue to discuss this feature and, if accepeted, eventually use a dedicated PR for it.

silkrpc/context_pool.hpp
canepat: Good point. I think `select_context` or `next_context` are better names than simply `get_context`

silkrpc/context_pool.hpp
canepat: I kindly disagree on this.

The presence of two different methods comes exactly from two differen needs: sometimes clients need to get the next `Context` in the pool to access all its capabilities, but a lot of times they need just the next `asio::io_context` in the pool.

If we eliminate the second method, a chained call `get_context().io_context()` will be present in many call sites and I think that would just make the code more verbose.

silkrpc/main.cpp
canepat: Agreed. I would leave this for a dedicated PR, you could open an issue just for tracking

lupin012: add issue #280

silkrpc/context_pool.cpp
lupin012: Created issue #281

PR #279, opened 2022-05-14
Forkchoice fix

enriavil1:


silkrpc/commands/engine_api.cpp
```diff
@@ -94,7 +97,23 @@ asio::awaitable<void> EngineRpcApi::handle_engine_forkchoice_updated_v1(const nl
     #ifndef BUILD_COVERAGE
     try {
     #endif
+        evmc::bytes32 empty_hash = 0x0000000000000000000000000000000000000000000000000000000000000000_bytes32;
```
canepat, line 100: You can use `constexpr auto zero_hash = 0_bytes32;` to obtain a shorter version of this line.

Moreover, `zero_hash` seems to be a better name here because `empty_hash` is more suitable to indicate the hash of the empty ("") string

silkrpc/commands/engine_api.cpp
```diff
@@ -94,7 +97,23 @@ asio::awaitable<void> EngineRpcApi::handle_engine_forkchoice_updated_v1(const nl
...
         const ForkchoiceState forkchoice_state = params[0].get<ForkchoiceState>();
+
+        if (forkchoice_state.safe_block_hash == empty_hash) {
+            auto error_msg = "safe block hash is empty";
```
canepat, line 104: May `error_msg` be `const` here?

silkrpc/commands/engine_api.cpp
```diff
@@ -94,7 +97,23 @@ asio::awaitable<void> EngineRpcApi::handle_engine_forkchoice_updated_v1(const nl
...
+            SILKRPC_ERROR << error_msg << "\n";
+            reply = make_json_error(request.at("id"), 100, error_msg);
+            co_return;
+        }
+
+        if (forkchoice_state.finalized_block_hash == empty_hash) {
+            auto error_msg = "finalized block hash is empty";
```
canepat, line 111: May `error_msg` be `const` here?

canepat (changes requested):

PR #287, opened 2022-05-20
better not implemented method message

enriavil1:
Added method to not implemented method message so we include the missing method inside of our message


canepat (changes requested):
I very much appreciate this tiny PR for an improved error message. Thanks!

Please add one integration test using some invalid method (e.g. "unknown_method")

canepat (approved):

Issue #204, opened 2021-12-27
Explore usage of asio-grpc library


canepat:
Explore usage of [asio-grpc](https://github.com/Tradias/asio-grpc) library

This could help simplifying the code and hopefully improving the performance too

Tradias:
Heya, I am the author of asio-grpc. Is there anything I can help you with here? Do you need any information regarding the library?

canepat:
Hey @Tradias, thanks a lot for asking! As you can see, this issue is a reminder for a future research work using your library.

Currently we use `asio::awaitable` as C++ coroutine awaitable type and we have some code wrapping the async gRPC API into coroutines. However, such mapping occurs at the application level and just for our specific RPCs, while using asio-grpc it would happen inside the library at the infrastructure level and so we could get rid of some code not so . Moreover, your idea to wrap `asio::execution_context` around `grpc::CompletionQueue` seems really interesting to improve our threading model.

I will be back to you asking for some help when we will start working on this.

Tradias:
Fyi, I have started to work on a pull request that shows what code simplification and correctness asio-grpc would bring.

canepat:
> Fyi, I have started to work on a pull request that shows what code simplification and correctness asio-grpc would bring.

Thanks a lot, @Tradias! Looking forward to see your PR, please let me know if you have questions or any issue with our code.

Tradias:
Looks like this is done now :).

canepat:
Closed by #326

PR #304, opened 2022-06-11
asio-grpc

Tradias:
https://github.com/torquem-ch/silkrpc/issues/204

I adjusted only the EthBackend to act as an example and reference for feedback. None of the other tests or main functions are expected to work.

Your feedback is now needed to decide whether the direction shown in this pr suits your plans. Also let me know whether I should refactor the streaming RPC as well, so you can get an idea on how it might look like.

It seems that you are not using clang-format for your code, so I did not know how to format my changes and kept them mostly ugly.

Highlights:

* Simplified protoc invocation through asio-grpc's `asio_grpc_protobuf_generate`
* Easier declaration of unary RPCs (no more `using XX = AsyncUnaryClient<...>`)
* Optionally switching to a different execution context upon completion of the RPC. Not sure whether it tis needed, but since the previous code was doing it, I put it in.
* Mocked stub in tests as an example to show how to simplify testing and make it more strict. Also makes it easier to run tests in parallel since no search for a free port is necessary.

---

There are a few local changes that I made to get things to compile:

* Add an include to `<utility>` in all places that use `asio::awaitable`. I think this only affects GCC12 and Boost.Asio pre 1.79.
* In `silkworm/cmake/Hunter/config.cmake` remove URL+SHA1 from re2 and CLI11.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,8 +30,8 @@ include(silkworm/third_party/evmone/cmake/cable/bootstrap.cmake)
 include(CableBuildInfo)
 include(silkworm/third_party/evmone/cmake/cable/HunterGate.cmake)
 HunterGate(
-  URL "https://github.com/cpp-pm/hunter/archive/v0.24.1.tar.gz"
-  SHA1 "4942227a6e6f5e64414c55b97ef98609de199d18"
+  URL "https://github.com/cpp-pm/hunter/archive/v0.24.3.tar.gz"
+  SHA1 "10738b59e539818a01090e64c2d09896247530c7"
   FILEPATH "${CMAKE_SOURCE_DIR}/cmake/Hunter/config.cmake"
 )
 
--- a/cmake/Hunter/config.cmake
+++ b/cmake/Hunter/config.cmake
@@ -16,3 +16,10 @@
 
 # silkworm configuration
 include(${CMAKE_SOURCE_DIR}/silkworm/cmake/Hunter/config.cmake)
+
+hunter_config(
+    asio-grpc
+    VERSION ${HUNTER_asio-grpc_VERSION}
+    CMAKE_ARGS
+      ASIO_GRPC_USE_BOOST_CONTAINER=ON
+)
--- a/cmake/Hunter/packages.cmake
+++ b/cmake/Hunter/packages.cmake
@@ -15,7 +15,6 @@
 ]]
 
 hunter_add_package(abseil)
-hunter_add_package(asio)
 hunter_add_package(Catch)
 hunter_add_package(ethash)
 hunter_add_package(gRPC)
@@ -23,3 +22,4 @@ hunter_add_package(GTest)
 hunter_add_package(intx)
 hunter_add_package(nlohmann_json)
 hunter_add_package(Protobuf)
+hunter_add_package(asio-grpc)
--- a/cmd/ethbackend_coroutines.cpp
+++ b/cmd/ethbackend_coroutines.cpp
@@ -63,7 +63,7 @@ int ethbackend_coroutines(const std::string& target) {
         silkrpc::ContextPool context_pool{1, create_channel};
         auto& context = context_pool.next_context();
         auto io_context = context.io_context();
-        auto grpc_queue = context.grpc_queue();
+        auto grpc_context = context.grpc_context();
 
         boost::asio::signal_set signals(*io_context, SIGINT, SIGTERM);
         signals.async_wait([&](const boost::system::error_code& error, int signal_number) {
@@ -74,7 +74,7 @@ int ethbackend_coroutines(const std::string& target) {
         const auto channel = grpc::CreateChannel(target, grpc::InsecureChannelCredentials());
 
         // Etherbase
-        silkrpc::ethbackend::BackEndGrpc eth_backend{*io_context, channel, grpc_queue};
+        silkrpc::ethbackend::BackEndGrpc eth_backend{*io_context, channel, *grpc_context};
         boost::asio::co_spawn(*io_context, ethbackend_etherbase(eth_backend), [&](std::exception_ptr exptr) {
             context_pool.stop();
         });
--- a/cmd/unit_test.cpp
+++ b/cmd/unit_test.cpp
@@ -14,5 +14,37 @@
    limitations under the License.
 */
 
-#define CATCH_CONFIG_MAIN
+#include <gmock/gmock.h>
+
+#include <string>
+
+#define CATCH_CONFIG_RUNNER
 #include <catch2/catch.hpp>
+
+int main(int argc, char** argv) {
+    ::testing::InitGoogleMock(&argc, argv);
+
+    auto& listeners = ::testing::UnitTest::GetInstance()->listeners();
+    delete listeners.Release(listeners.default_result_printer());
+
+    class CatchListener : public ::testing::EmptyTestEventListener {
+      public:
+        virtual void OnTestPartResult(const ::testing::TestPartResult& result) override {
+            if (!result.failed()) {
+                return;
+            }
+            const auto file = result.file_name() ? result.file_name() : "unknown";
+            const auto line = result.line_number() != -1 ? result.line_number() : 0;
+            const auto message = result.message() ? result.message() : "no message";
+            ::Catch::SourceLineInfo source_line_info(file, static_cast<std::size_t>(line));
+            auto result_disposition = result.fatally_failed() ? ::Catch::ResultDisposition::Normal
+                                                              : ::Catch::ResultDisposition::ContinueOnFailure;
+            ::Catch::AssertionHandler assertion("GTEST", source_line_info, "", result_disposition);
+            assertion.handleMessage(::Catch::ResultWas::ExplicitFailure, result.message());
+            assertion.setCompleted();
+        }
+    };
+    listeners.Append(new CatchListener);
+
+    return Catch::Session().run(argc, argv);
+}
--- a/silkrpc/CMakeLists.txt
+++ b/silkrpc/CMakeLists.txt
@@ -20,11 +20,9 @@ find_package(intx CONFIG REQUIRED)
 # Find Protobuf installation
 set(protobuf_MODULE_COMPATIBLE TRUE)
 find_package(Protobuf CONFIG REQUIRED)
-find_program(PROTOBUF_PROTOC protoc REQUIRED)
 
-# Find gRPC installation
-find_package(gRPC CONFIG REQUIRED)
-find_program(GRPC_CPP_PLUGIN_EXECUTABLE grpc_cpp_plugin REQUIRED)
+# Find asio-grpc installation
+find_package(asio-grpc CONFIG REQUIRED)
 find_package(mimalloc 2.0 REQUIRED)
 
 # Define gRPC proto files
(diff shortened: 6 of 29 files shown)


canepat:
> I adjusted only the EthBackend to act as an example and reference for feedback. None of the other tests or main functions are expected to work.
> 
@Tradias first of all, thanks a lot!

> Your feedback is now needed to decide whether the direction shown in this pr suits your plans. Also let me know whether I should refactor the streaming RPC as well, so you can get an idea on how it might look like.
>

The direction you're showing in this PR is definitively interesting: the resulting code is nice, terse and clear, so it looks very promising. I need to jump on this after finishing current work on state cache and take a closer look to give additional feedback.

If you could also refactor the bidirectional-streaming RPC, it would be really great!
 
> It seems that you are not using clang-format for your code, so I did not know how to format my changes and kept them mostly ugly.
> 
Yes, my bad. We should be more consistent as far as formatting is concerned.

Tradias:
I adjusted the RemoteCursor and RemoteTransaction to asio-grpc, including one or two of their unit tests. This time I didn't do the executor switch. The gRPC API (and thereby also asio-grpc) is thread-safe, but RemoteTransaction is not thread-safe, so that needs to be addressed. Here it might be beneficial to treat the GrpcContext as an execution_context and use `co_await asio::dispatch(asio::bind_executor(grpc_context, asio::use_awaitable))` before accessing the std::map stored in the remote transaction.

A few more things that I noticed:

* GMock was not hooked into the catch2 runner, which means any failed assertion in GMock were not marking a test as failed
* There was no place for tests to share code, so I added a subdirectory `silkrpc/test` for that

Tradias:
@canepat Did you find some time to look at the streaming RPC code: https://github.com/torquem-ch/silkrpc/pull/304/commits/b68b095c074fd8e0977b26dddbf0fe19fa8502d0

What are the next steps now?

canepat:
> @canepat Did you find some time to look at the streaming RPC code: [b68b095](https://github.com/torquem-ch/silkrpc/commit/b68b095c074fd8e0977b26dddbf0fe19fa8502d0)
> 
@Tradias TBH I've just started working on it yesterday evening :smile:

> What are the next steps now?
Thanks for asking. I need some time to understand and apply the changes to the latest code, then I will give you a feedback and maybe ask for your help. When finished this, next step is a complete functional and performance regression test.

I've started my work on a local branch of mine different from yours, mainly because we're using standalone Asio in master.

Tradias:
Great news.

I have based this branch on https://github.com/torquem-ch/silkrpc/pull/285 because I thought that would be the direction you want to go. I have also updated standalone Asio on Hunter to its latest version in release [0.24.3](https://github.com/cpp-pm/hunter/releases/tag/v0.24.3) in case you need that.

I would also add a fix for the potential race condition in RemoteTransaction.

canepat:
> I adjusted the RemoteCursor and RemoteTransaction to asio-grpc, including one or two of their unit tests. This time I didn't do the executor switch. The gRPC API (and thereby also asio-grpc) is thread-safe, but RemoteTransaction is not thread-safe, so that needs to be addressed. Here it might be beneficial to treat the GrpcContext as an execution_context and use `co_await asio::dispatch(asio::bind_executor(grpc_context, asio::use_awaitable))` before accessing the std::map stored in the remote transaction.
> 
In the current design the RemoteTransaction does not need to be thread-safe because it is always used by the same `asio::io_context` thread contained in `silkrpc::Context`: each `silkrpc::Context` has its own `asio::io_context` and its own `ethdb::kv::RemoteDatabase`, a factory of `ethdb::kv::RemoteTransaction` instances accessed only by such `asio::io_context`. This design has a two-fold advantage:
- once the Asio <-> gRPC coroutine glue has been implemented and the `silkrpc::Context`-based threading model established, you should not worry about threading anymore, just be aware not to do any blocking operation in your business logic
- you can configure N execution contexts in `silkrpc::ContextPool` obtaining N parallel cooperative schedulers, each of them running its own coroutine frame. They share some state at the `silkrpc::Context` level (namely, `BlockCache` and [recently](https://github.com/torquem-ch/silkrpc/pull/318/files) also `StateCache`) and there synchronization takes place, but just there.

At the gRPC frontier, this is achieved by posting the completion handler of any async gRPC operation to the `asio::io_context` that started such operation (is this the executor switch you're referring to? I think so). At least this was the intention, bugs apart :smile: Please let me know if this explains how it works now and if you see any fallacies or possible improvements in this setup. 

Hence, I expect that we should achieve the same setup also when using `asio-grpc`. To my mind the potential race condition in `ethdb::kv::RemoteTransaction` is maybe due to the lack of executor switch in async operation completion. but I could be missing something.

> A few more things that I noticed:
> 
> * GMock was not hooked into the catch2 runner, which means any failed assertion in GMock were not marking a test as failed
> * There was no place for tests to share code, so I added a subdirectory `silkrpc/test` for that

Thanks a lot for these improvements!

canepat:
> I have based this branch on #285 because I thought that would be the direction you want to go.
Yes, I tried to switch from `Asio` to `Boost.Asio` using the versions included in Hunter [0.24.1](https://github.com/cpp-pm/hunter/releases/tag/v0.24.1) but during our performance workload I saw Silkrpc crashing repeatedly at high rates (e.g. qps in the thousands), so I decided to pause the activity.

> I have also updated standalone Asio on Hunter to its latest version in release [0.24.3](https://github.com/cpp-pm/hunter/releases/tag/v0.24.3) in case you need that.
> 

This is good news, thank you. I will give it a try.

Tradias:
Yes I understand, you would like to run one io_context+GrpcContext per thread and then process a request or sequence of requests on exactly one such pair. The dream scenario 😃 yet I haven't found a perfect way of implementing it. There are always tradeoffs, as they say.

By executor switching I mean switching between the io_context and GrpcContext within one pair. If they both run on the same thread then the switching is unnecessary. Nonetheless, if the switching is implemented correctly, e.g. by using [asio::dispatch](https://www.boost.org/doc/libs/1_79_0/doc/html/boost_asio/reference/dispatch/overload1.html). then it is essentially free if the contexts run on the same thread (just one thread_local pointer comparison), while the code remains compatible with a two-thread model. I suspect however that the code will naturally evolve in such that way that it assumes that the io_context and GrpcContext are running on the same thread, making a switch back to the two-thread model more difficult.

Now to the tradeoffs in the current implementation of running an io_context and GrpcContext in the same thread. Well, under load its all fine, just .poll() the contexts in turn. During idle time I let the GrpcContext sleep in a call to [AsyncNext](https://grpc.github.io/grpc/cpp/classgrpc_1_1_completion_queue.html#a290ddf3d96e6d1d5886580e60b0602b4) for a few microseconds. That works, but means that the io_context might have some time latency when going from idle to loaded. The implementation can be seen [here](https://github.com/Tradias/asio-grpc/blob/master/src/agrpc/run.hpp#L138-L162).

canepat:
@Tradias here it is the branch where I've applied your changes on top of the latest code: [asio_grpc_integration](https://github.com/torquem-ch/silkrpc/tree/asio_grpc_integration)

As you can see from CI such branch compiles and also seems to pass tests, *but* sometimes I've encountered unpredictable segmentation faults during the execution of unit tests on my local machine. I've struggled some days to debug the problem but now I think I understand what happens.

__Description__
Basically the problem appears during the execution of `RemoteDatabase`/`RemoteTransaction2`/`RemoteCursor2` unit tests in https://github.com/torquem-ch/silkrpc/blob/asio_grpc_integration/silkrpc/test/kv_test_base.hpp#L34: after the `io_context` thread [the one running the coroutine frame stack after `spawn_and_wait`] calls the `BidiStreamingRpc::request_and_read` and eventually hits the `agrpc::process_grpc_tag` function in `expect_request_async_tx`, there is a potential race between such thread and the `GrpcContext` thread [the  one executing the `run_completion_queue` loop].
Depending on the thread scheduling, the following unlikely scenario can happen:
- the `io_context` thread is suspended *just after* the `agrpc::process_grpc_tag` call but *before* extracting and returning the `reader_writer_ptr_` raw pointer, which thus has not been assigned to the `reader_writer_` unique pointer within the enclosing `BidiStreamingRpc` instance;
- the `GrpcContext` thread is immediately resumed by the `grpc::Alarm` expiration triggered by `agrpc::process_grpc_tag` and executes one cycle of the `run_completion_queue` loop by calling the completion handler for the `agrpc::request` async function, which is `BidiStreamingRpc::ReadNext::operator(Op& op, bool ok)` where the subsequent `agrpc::read` call is fired passing an empty `reader_writer_` unique pointer into it.

This is the root cause of the segmentation violation error, which then happens in `agrpc/detail/rpc.hpp::61` executing this functor:
```
void operator()(const agrpc::GrpcContext&, void* tag) { responder.Read(&message, tag); }
```
where the `responder` reference is zero, but just because it has been unwrapped from an empty unique pointer in the first place.

__Analysis__
I think this problem is not limited to this somewhat artificial test setup, but it's also possibile in production code. I had a vague recall to have encountered similar issues with my original code as well many months ago, so I've just searched for the relevant issue in gRPC C++ library and found again this comment:

https://github.com/grpc/grpc/issues/13019#issuecomment-337977032

If you use multi-threaded CQ [as it is in our case] and call `Async...` generated methods in client stubs API [as `asio-grpc` does], you basically end up having a potential race condition unless you use locks. Using the two `PrepareAsync...` + `StartCall` async API to start any RPC on async client is the recommended solution and I can confirm that I didn't have any more concurrency issues after switching to such 2-call API.

__Solution__
I managed to solve the issue in unit tests by wrapping the `agrpc::process_grpc_tag` call into an enclosing `asio::post` on the `io_context` instance, but this is just a trick I did to confirm the diagnosis of the problem. In my opinion the correct solution is adding support for `PrepareAsync...` + `StartCall` async API for client-side generated stubs in `agrpc/detail/RequestFn`, if possible.

@Tradias what do you think? are you willing to introduce such feature in `asio-grpc`?

Tradias:
Ah good news. Sorry that you had to go through that debugging while I had already fix it in this commit https://github.com/torquem-ch/silkrpc/pull/304/commits/ecbb86741aaba733c8f583dab4b5854447df929a ^^.

I see, using `PrepareAsync` makes it easier to mock the call in testing. I will take a look at it in asio-grpc.

canepat:
> Ah good news. Sorry that you had to go through that debugging while I had already fixed it in this commit [ecbb867](https://github.com/torquem-ch/silkrpc/commit/ecbb86741aaba733c8f583dab4b5854447df929a) ^^.
> 
Oh, it seems that I started from the previous commit probably a few hours before your last one :facepalm: 

> I see, using `PrepareAsync` makes it easier to mock the call in testing.
>
Sorry for asking again on this but I'm really interested in checking my understanding with you.

I think that using `PrepareAsync` + `StartCall` does not only make easier to mock the call in testing, but also solves a race condition in production code. In other words, currently `agrpc::request` API is not thread-safe to use in a multi-threaded scenario with generated client-side streaming stubs. Please let me know if I am wrong here.

> I will take a look at it in asio-grpc.
>
Thanks a lot!

Tradias:
If I read [this](https://github.com/grpc/grpc/pull/12269)

>This is important because it enables an object to be fully constructed and available before another thread sees its CQ event trigger.

 correctly then the race condition would only happen if one GrpcContext can be run on multiple thread simultaneously, which is not supported by asio-grpc.

canepat:
What do you mean exactly by

> if one GrpcContext can be run on multiple thread simultaneously

? In the following, I assume that you mean: multiple threads are calling `GrpcContext::run` or `GrpcContext::run_completion_queue` simultaneously.

My interpretation is that the "object to be fully constructed" in such sentence is exactly the gRPC ReaderWriter instance returned by the `Async` methods, so I think this is not related to multiple threads calling `GrpcContext::run`.

In fact, my code is calling `GrpcContext::run_completion_queue` with just one thread, whilst another one thread calls `io_context::run`. The race happens when the `io_context` thread calls `agrpc::request`, thus putting a tag on the `grpc::CompletionQueue` before gRPC ReaderWriter object is fully constructed and opening the door for a potential race condition in the completion handler called by the `GrpcContext` thread.

Tradias:
You might be right. Although since `agrpc::request` finishes on the thread of the GrpcContext one would have to switch to the thread of the io_context first before interacting with the ReaderWriter to run into the race-condition. Nonetheless, it is easy to fix with the `PrepareAsync...` overloads which I am adding atm: https://github.com/Tradias/asio-grpc/issues/33

Tradias:
I have implemented support for PrepareAsync in asio-grpc, it will be part of the next release (2.0.0). You can see what the changes in silkrpc look like in this commit https://github.com/torquem-ch/silkrpc/pull/304/commits/4232a569e580bc81c82baad4c73915a2a1cf2911.

Are there any more issues or wishes that you encountered while working on https://github.com/torquem-ch/silkrpc/pull/326? I am happy to address them before releasing asio-grpc 2.0.0.

canepat:
> I have implemented support for PrepareAsync in asio-grpc, it will be part of the next release (2.0.0). You can see what the changes in silkrpc look like in this commit [4232a56](https://github.com/torquem-ch/silkrpc/commit/4232a569e580bc81c82baad4c73915a2a1cf2911).
> 

Thank you very much! Also for the usage examples :pray: Looking forward to see the next release included in Hunter.

> Are there any more issues or wishes that you encountered while working on #326? I am happy to address them before releasing asio-grpc 2.0.0.

A couple of ideas:
- it could be useful to directly include in asio-grpc the support for dispatching the completion handlers either inline or in a specified executor (`InlineDispatcher`/`ExecutorDispatcher`) as you did [here](https://github.com/torquem-ch/silkrpc/pull/304/files#diff-637cf42b866b92649451ee9bf03a5a5dfbc839614628f67f323ca606720c7f0cR41)
- you could decide to go even further and include base implementations for unary and client/server/bidirectional streaming RPC classes both client-side and maybe also server-side. I suppose that similar (if not identical classes) are written in any code using asio-grpc for production. Surely some use cases can be specific (e.g. in bidirectional streaming having tic-toc write/read alternation is just one specific use case), but both single and compound methods could be included to cover all the needs or just offer the basic primitives and let users extend them.

Just my 2 cents.

Tradias:
Thanks for your feedback! I am thinking of adding more high-level functionality to asio-grpc and your input is exactly what I need for that.

Regarding the dispatcher, it is already possible in asio-grpc using GrpcStream, e.g. the following code

```c++
agrpc::GrpcStream stream{grpc_context};
co_await stream.initiate(agrpc::write, request).next(asio::bind_executor(io_context, asio::use_awaitable)); 
```

will finish the `write` and then dispatch to the `io_context`. It should also be possible to create a CompletionToken decorator like `asio::bind_executor`, I can give you an example later. I am surprised you implemented the dispatching again for the server-streaming RPCs, I thought you want to run the io_context and grpc_context on the same thread? Btw for that there is `agrpc::run` and a new `agrpc::run_completion_queue`, see https://github.com/Tradias/asio-grpc/blob/c4bf4b277232bc20df39efacaf7d0fb8314d19b7/example/share-io-context-client.cpp#L87-L98.

Adding compound method sounds like a good idea. If you are curious, here is another project that uses asio-grpc and we can indeed see the combined [request+finish for unary RPCs](https://github.com/meta-soul/MetaSpore/blob/bd7ac83abeceb7422b01883e8fea2833fcbc84ab/cpp/serving/py_preprocessing_model.cpp#L133-L134). And for [server-side unary RPCs](https://github.com/meta-soul/MetaSpore/blob/bd7ac83abeceb7422b01883e8fea2833fcbc84ab/cpp/serving/grpc_server.cpp#L125-L150) they create some kind of `expected<Response, grpc::Status>` and then finish the RPC. That could be offered by asio-grpc as well, e.g.

```c++
agrpc::repeatedly_request_unary(&Load::AsyncService::RequestLoad, load_service, 
[&](grpc::ServerContext &ctx, LoadRequest &req) -> asio::awaitable<std::expected<LoadReply, grpc::Status>>
{
  // create reply and co_return it, asio-grpc will then `agrpc::finish(writer, reply, grpc::Status::OK)`
  // or return a non-ok Status and asio-grpc will `agrpc::finish_with_error(writer, status)`
});
```

Tradias:
Here is the CompletionExecutorBinder example: https://github.com/Tradias/example-vcpkg-grpc/blob/asio-grpc-bind-completion-executor/bind_completion_executor.hpp. It could be performance optimized a little bit more, currently the work guard is being recreated instead of moved. It makes heavy use of C++20 including CTAD for aggregates, not sure whether that is supported by clang, but it can easily be backported to C++17 if needed.

You could use in silkrpc either here:

```c++
    template<typename Executor, typename CompletionToken = agrpc::DefaultCompletionToken>
    auto finish_on(const Executor& executor, const Request& request, CompletionToken&& token = {}) {
        return boost::asio::async_compose<CompletionToken, void(boost::system::error_code, Reply)>(
            Call/*no longer need detail::ExecutorDispatcher<Executor>*/{*this, request, executor}, 
            agrpc_ext::CompletionExecutorBinder{executor, token}, 
            executor);
    }
```

Or even on the caller side:

```c++
const auto reply = co_await eb_rpc.finish(::remote::EtherbaseRequest{}, 
                                          agrpc_ext::CompletionExecutorBinder{executor_, asio::use_awaitable});
```

Tradias:
@canepat Asio-grpc v2.0.0 update is pending here: https://github.com/cpp-pm/hunter/pull/593

canepat:
> I am thinking of adding more high-level functionality to asio-grpc and your input is exactly what I need for that.
> 

More high-level functions in asio-grpc sounds great!

> Regarding the dispatcher, it is already possible in asio-grpc using GrpcStream, e.g. the following code
> 
> ```c++
> agrpc::GrpcStream stream{grpc_context};
> co_await stream.initiate(agrpc::write, request).next(asio::bind_executor(io_context, asio::use_awaitable)); 
> ```
> 
> will finish the `write` and then dispatch to the `io_context`.

I didn't know, thanks for pointing this out.

> I am surprised you implemented the dispatching again for the server-streaming RPCs, I thought you want to run the io_context and grpc_context on the same thread?
> 

I did it because currently in Silkrpc we can choose w/ startup option either single-threaded or multi-threaded execution loop: in the latter case, io_context and grpc_context run on different threads with grpc_context thread used only to wait and immediately dispatch the incoming events from the `grpc::CompletionQueue` to the io_context thread (so that in both cases all the business logic does not need to be thread-safe).
Hence, in the single-threaded scenario I need an inline dispatching (no thread switch), in the multi-threaded one a real dispatching from grpc_context to io_context.

> Btw for that there is `agrpc::run` and a new `agrpc::run_completion_queue`, see https://github.com/Tradias/asio-grpc/blob/c4bf4b277232bc20df39efacaf7d0fb8314d19b7/example/share-io-context-client.cpp#L87-L98.
> 

Currently I'm using `GrpcContext::poll_completion_queue` in the single-threaded execution loop because when neither io_context nor grpc_context have done any work, we adopt different idle wait strategies to balance CPU waste and latency and AFAIK `agrpc::run` does not allow to specify how to handle idle cycles. IMHO this could be another interesting feature.

> Adding compound method sounds like a good idea. If you are curious, here is another project that uses asio-grpc and we can indeed see the combined [request+finish for unary RPCs](https://github.com/meta-soul/MetaSpore/blob/bd7ac83abeceb7422b01883e8fea2833fcbc84ab/cpp/serving/py_preprocessing_model.cpp#L133-L134). And for [server-side unary RPCs](https://github.com/meta-soul/MetaSpore/blob/bd7ac83abeceb7422b01883e8fea2833fcbc84ab/cpp/serving/grpc_server.cpp#L125-L150) they create some kind of `expected<Response, grpc::Status>` and then finish the RPC. That could be offered by asio-grpc as well, e.g.
> 
> ```c++
> agrpc::repeatedly_request_unary(&Load::AsyncService::RequestLoad, load_service, 
> [&](grpc::ServerContext &ctx, LoadRequest &req) -> asio::awaitable<std::expected<LoadReply, grpc::Status>>
> {
>   // create reply and co_return it, asio-grpc will then `agrpc::finish(writer, reply, grpc::Status::OK)`
>   // or return a non-ok Status and asio-grpc will `agrpc::finish_with_error(writer, status)`
> });
> ```

Thanks for sharing, I will take a look.

canepat:
> Here is the CompletionExecutorBinder example: https://github.com/Tradias/example-vcpkg-grpc/blob/asio-grpc-bind-completion-executor/bind_completion_executor.hpp. It could be performance optimized a little bit more, currently the work guard is being recreated instead of moved. It makes heavy use of C++20 including CTAD for aggregates, not sure whether that is supported by clang, but it can easily be backported to C++17 if needed.
> 

Thanks again also for the usage example in Silkrpc. `CompletionExecutorBinder` will not be available in asio-grpc 2.0.0 though, correct? Are you considering to add it in the future?

Tradias:
> I did it because currently in Silkrpc we can choose w/ startup option either single-threaded or multi-threaded execution loop

I understand, the good thing is in both cases you can use the same code (the one that dispatches to the io_context) since it will automatically become a no-op when the io_context is already running in the current thread.

> Currently I'm using GrpcContext::poll_completion_queue in the single-threaded execution loop because when neither io_context nor grpc_context have done any work, we adopt different idle wait strategies to balance CPU waste and latency and AFAIK agrpc::run does not allow to specify how to handle idle cycles. IMHO this could be another interesting feature.

I have been fiddling with different wait strategies as well and if you find a better one then let me know :), because that is exactly what `agrpc::run` is supposed to take care of for the user. I think the one I use now is the best possible though, which is to sleep in [grpc::CompletionQueue::AsyncNext](https://grpc.github.io/grpc/cpp/classgrpc_1_1_completion_queue.html#a290ddf3d96e6d1d5886580e60b0602b4), because in that case there is zero latency for gRPC operations. Alternatively, it would also be possible to sleep in [io_context::run_one_until](https://www.boost.org/doc/libs/1_79_0/doc/html/boost_asio/reference/io_context/run_one_until.html). I can see making that a configuration option, but that is about it, the wait strategies that you have implemented are not as performant as those two and will therefore not be implemented in asio-grpc. So please test `agrpc::run_completion_queue` and let me know how it performs for you. Note that I also try to benchmark it from a server perspective with an idle io_context: https://github.com/Tradias/asio-grpc#performance (it is called cpp_asio_grpc_io_context_coro).

> Thanks again also for the usage example in Silkrpc. CompletionExecutorBinder will not be available in asio-grpc 2.0.0 though, correct? Are you considering to add it in the future?

No, the high-level API will follow Asio's logic of IO-Objects which already support binding a completion executor using the normal `asio::bind_executor` similar to how I showed it for `agrpc::GrpcStream`. I guess I can imagine something like:

```c++
agrpc::GrpcStub grpc_stub{stub, grpc_context};
grpc::ClientContext client_context;
Request request;
boost::outcome<Response, grpc::Status> response = 
   co_await grpc_stub.unary(&Stub::AsyncUnary, client_context, request, asio::bind_executor(io_context, asio::use_awaitable));
```

Tradias:
I have started to design the high-level client-side API, if you are curious take a look at: https://github.com/Tradias/asio-grpc/issues/35. Your input is of course very welcome.

canepat:
I'm closing this because #326, which supersedes it, has been successfully merged.

@Tradias thanks again for your help!

PR #314, opened 2022-06-21
added parity list storage keys

enriavil1:


silkrpc/commands/parity_api.cpp
```diff
@@ -77,4 +80,70 @@ asio::awaitable<void> ParityRpcApi::handle_parity_get_block_receipts(const nlohm
...
+        // we look for keys until we have the quantity we want or the key is invalid/empty
+        while (v.size() >= silkworm::kHashLength && keys.size() != quantity) {
+            keys.push_back(v.substr(0, silkworm::kHashLength));
+            const auto kv_pair = co_await cursor->next();
+            if (kv_pair.key != seek_bytes) {
+                break;
+            }
+            v = kv_pair.value;
+        }
+        co_await cursor->close_cursor();
```
canepat, line 134: It is good practice to handle resources' lifetime carefully, so it is definitively correct here closing the cursor when done.

In this specific case, however, you should _remove_ this call :smile: because remote cursors created within a remote transaction will be automatically closed on server-side when the remote transaction is closed (i.e. `co_await tx->close();` at line 146).

canepat (changes requested):

canepat (approved):

PR #320, opened 2022-06-28
Integration test rework

lupin012:


tests/integration/run_tests.py
```diff
@@ -0,0 +1,215 @@
...
+    """ parse command line and execute tests
+    """
+    exit_on_fail = 1
+    silk = 1
+    loop_number = 1
+    verbose = 0
+    req_test = -1
+    api_name = ""
+    verify_with_rpc = 0
+    json_dir = "./goerly/"
```
canepat, line 156: Maybe the chain name could be a command-line option to support more than one chain. It's OK to do it in another PR

tests/integration/run_tests.py
```diff
@@ -0,0 +1,215 @@
...
+    output_dir = "./int_test_results/"
```
canepat, line 157: I would put it within the chain folder, i.e. `./goerly/results`

.gitignore
```diff
@@ -6,4 +6,5 @@ perf.data*
 .vs
 *.code-workspace
 *.hrd
-*.stdout
\ No newline at end of file
+*.stdout
+int_test_results
```
canepat, line 10: `int_test_results` -> `results`

canepat (changes requested):

PR #323, opened 2022-07-02
Trace call int test

lupin012:
This PR contains:
- one FIX due to python indentation
- 3 more integration test on trace_call
- add -o option 
          The simple way to create integration tests is:
                   - create on  goerly/<api>/test_x.json filling the request with empty response
                   - makes run_test -t < test_number> -d compare the result with rpcdaemon 
                      (see on goerly/results/<api>/test... both output ) until are the same result
                   - makes run_test -t < test_number> -d -o capture in goerly/results/<api>/test... the output 
                   - copy in goerly/<api>/test_x.json the response


canepat (approved):
LGTM

PR #325, opened 2022-07-04
trace_replayBlockTransactions implementation

Sixtysixter:
Implementation of API  trace_replayBlockTransactions

### DIFF
--- a/silkrpc/commands/trace_api.cpp
+++ b/silkrpc/commands/trace_api.cpp
@@ -130,12 +130,41 @@ asio::awaitable<void> TraceRpcApi::handle_trace_raw_transaction(const nlohmann::
 
 // https://eth.wiki/json-rpc/API#trace_replayblocktransactions
 asio::awaitable<void> TraceRpcApi::handle_trace_replay_block_transactions(const nlohmann::json& request, nlohmann::json& reply) {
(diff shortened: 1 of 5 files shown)


silkrpc/commands/trace_api.cpp
```diff
@@ -130,12 +130,41 @@ asio::awaitable<void> TraceRpcApi::handle_trace_raw_transaction(const nlohmann::
...
+        if (entry == "trace") {
+            config.trace = true;
+        }
+        if (entry == "vmTrace") {
+            config.vm_trace = true;
+        }
+        if (entry == "stateDiff") {
+            config.state_diff = true;
+        }
+    }
```
canepat, line 154: I think that adding the definition of `from_json` function for the type `trace::TraceConfig` would allow to avoid this code repetition in all `trace_` API functions.

After such change the code should become: `const auto config = params[1].get<trace::TraceConfig>();`

canepat (changes requested):

silkrpc/commands/trace_api.cpp
Sixtysixter: Great suggestion!

canepat (approved):

PR #326, opened 2022-07-08
Refactoring to use asio-grpc

canepat:


silkrpc/grpc/unary_rpc.hpp
```diff
@@ -0,0 +1,115 @@
...
+class UnaryRpc;
+
+template<
+    typename Stub,
+    typename Request,
+    template<typename> typename Reader,
+    typename Reply,
+    std::unique_ptr<Reader<Reply>>(Stub::*Async)(grpc::ClientContext*, const Request&, grpc::CompletionQueue*)
+>
+class UnaryRpc<Async> {
```
Tradias, line 53: If we assume that the only completion token being used is `asio::use_awaitable` then this entire class can be replaced by:

```cpp
template <class RPC, class CompletionExecutor = asio::system_executor>
auto unary_rpc(agrpc::GrpcContext& grpc_context, typename RPC::Stub& stub, const typename RPC::Request& request,
               const CompletionExecutor& completion_executor = asio::system_executor{})
    -> asio::awaitable<typename RPC::Response>
{
    grpc::ClientContext context;
    typename RPC::Response response;
    auto rpc = co_await RPC::request(grpc_context, stub, context, request, response,
                                     asio::bind_executor(completion_executor, asio::use_awaitable));
    if (!rpc.ok())
    {
        // In your code you are first turning the grpc::Status to a asio::error_code which is then thrown by Asio. I
        // think it would be easier to just throw the desired exception type and message directly, example:
        throw std::runtime_error(rpc.error_message());
    }
    co_return response;
}
```

with asio-grpc's new high-level client API.

And on the call side we would write:

```cpp
using RPC = agrpc::RPC<&::remote::ETHBACKEND::StubInterface::AsyncEtherbase>;
const auto reply = co_await silkrpc::unary_rpc<RPC>(grpc_context_, *stub, /*::remote::EtherbaseRequest*/{}, executor_);
```

lupin012 (approved):

silkrpc/ethdb/kv/state_changes_stream.cpp
```diff
@@ -18,89 +18,103 @@
...
-        cancelled_ = true;
-        SILKRPC_INFO << "Registration for state changes cancelled\n";
+asio::awaitable<void> StateChangesStream::run() {
+    SILKRPC_TRACE << "StateChangesStream::run state stream START\n";
+
+    auto cancellation_slot = cancellation_signal_.slot();
+
+    bool cancelled{false};
+    while (!cancelled) {
+        StateChangesRpc state_changes_rpc{*stub_, grpc_context_};
```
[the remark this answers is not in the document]
canepat: Thanks for sharing!

Your high-level client _basic_ API seems pretty clear and polished to me.

I know that treating normal server-side RPC completion as operation aborted is quite ugly, but it is just a quick-and-dirty trick to make it work with my current API design waiting for your client _compound_ API :smile:

silkrpc/grpc/unary_rpc.hpp
canepat: In terms of high-level API design for asio-grpc, I suppose you are planning to support any completion token. For sure this function can be a convenient all-in-one for `asio::awaitable`.

Tradias: So you would expect that everything is heap allocated so that a user can write:

```cpp
using RPC = agrpc::RPC<&remote::ETHBACKEND::StubInterface::AsyncEtherbase>;
RPC rpc{grpc_context};   // owns ClientContext and RPC::Response
rpc.client_context().set_deadline();
// callback as an example needs the lifetime extension of ClientContext and Response in the rpc
// that's why heap allocation is required
rpc.request(stub, request, [](std::expected<RPC::Response, grpc::Status> response) {
});
```

Not sure whether I like that so much. I don't want the new API to be a performance burden for users of coroutines or sender/receiver. I could allocate space for the `Response` object within the operation state of the current RPC::request call, I guess I will simply add an overload to the current API for that.

canepat: I completely agree that new API should _not_ place a performance burden when it is not necessary, so an overload is exactly the solution I would suggest for `asio-grpc` proposal to be complete.
To be completely clear, I'm going to use the coroutine API in Silkrpc so my suggestion is just a hint to make asio-grpc as complete as possible because, well, I love asio-grpc :smile:

PR #335, opened 2022-08-12
trace_get implementation

Sixtysixter:


silkrpc/commands/trace_api.cpp
```diff
@@ -254,15 +255,49 @@ asio::awaitable<void> TraceRpcApi::handle_trace_filter(const nlohmann::json& req
...
+    }
+
+    auto tx = co_await database_->begin();
     try {
         ethdb::TransactionDatabase tx_database{*tx};
 
-        reply = make_json_error(request["id"], 500, "not yet implemented");
+        const auto tx_with_block = co_await core::read_transaction_by_hash(*context_.block_cache(), tx_database, transaction_hash);
+        if (!tx_with_block) {
+            reply = make_json_content(request["id"]);
```
canepat, line 286: When block is not found the response is simply empty, instead of carrying some error (e.g. invalid transaction hash). Is this for compatibility with Parity?

silkrpc/commands/trace_api.cpp
```diff
@@ -254,15 +255,49 @@ asio::awaitable<void> TraceRpcApi::handle_trace_filter(const nlohmann::json& req
...
+            if (result.size() > index) {
+                reply = make_json_content(request["id"], result[index]);
+            } else {
+                reply = make_json_content(request["id"]);
+            }
+        }
     } catch (const std::exception& e) {
-        SILKRPC_ERROR << "exception: " << e.what() << " processing request: " << request.dump() << "\n";
-        reply = make_json_error(request["id"], 100, e.what());
+        reply = make_json_content(request["id"]);
```
canepat, line 300: When something goes wrong during execution the response is simply empty, instead of carrying some error. Again, is this for compatibility with Parity?

canepat (changes requested):

silkrpc/commands/trace_api.cpp
Sixtysixter: Yes. It is the same behavior of RPCDaemon: when an invalid hash is provided the result is simply empty.
Please note the trace_transaction  is implemented in the same way.

silkrpc/commands/trace_api.cpp
Sixtysixter: Yes. It is the same behavior of RPCDaemon: when an invalid hash is provided the result is simply empty.
Please note the trace_transaction  is implemented in the same way.
In addition note that when an invalid hash is provided an exception is raised in SilkRPC.

silkrpc/commands/trace_api.cpp
```diff
@@ -254,15 +255,49 @@ asio::awaitable<void> TraceRpcApi::handle_trace_filter(const nlohmann::json& req
...
     try {
         ethdb::TransactionDatabase tx_database{*tx};
 
-        reply = make_json_error(request["id"], 500, "not yet implemented");
+        const auto tx_with_block = co_await core::read_transaction_by_hash(*context_.block_cache(), tx_database, transaction_hash);
+        if (!tx_with_block) {
+            reply = make_json_content(request["id"]);
+        } else {
+            trace::TraceCallExecutor executor{*context_.io_context(), tx_database, workers_};
+            auto result = co_await executor.trace_transaction(tx_with_block->block_with_hash, tx_with_block->transaction);
```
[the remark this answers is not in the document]
Sixtysixter: Done

canepat (approved):

PR #342, opened 2022-08-21
Add trace_callMany API implementation

Sixtysixter:
trace_callMany API implementation


silkrpc/commands/trace_api.cpp
```diff
@@ -79,12 +79,33 @@ asio::awaitable<void> TraceRpcApi::handle_trace_call(const nlohmann::json& reque
 
 // https://eth.wiki/json-rpc/API#trace_callmany
 asio::awaitable<void> TraceRpcApi::handle_trace_call_many(const nlohmann::json& request, nlohmann::json& reply) {
+    auto params = request["params"];
```
canepat, line 82: `params` can be declared `const` here

silkrpc/commands/trace_api.cpp
```diff
@@ -79,12 +79,33 @@ asio::awaitable<void> TraceRpcApi::handle_trace_call(const nlohmann::json& reque
...
+    if (params.size() < 2) {
+        auto error_msg = "invalid trace_callMany params: " + params.dump();
+        SILKRPC_ERROR << error_msg << "\n";
+        reply = make_json_error(request["id"], 100, error_msg);
+        co_return;
+    }
+    const auto trace_calls = params[0].get<std::vector<trace::TraceCall>>();
+    const auto block_number_or_hash = params[1].get<BlockNumberOrHash>();
+
+    SILKRPC_INFO << "#trace_calls: " << trace_calls.size() << " block_number_or_hash: " << block_number_or_hash << "\n";
```
canepat, line 92: I would demote this log line to `DEBUG`

silkrpc/core/evm_trace.cpp
```diff
@@ -61,6 +61,13 @@ std::ostream& operator<<(std::ostream& out, const TraceConfig& tc) {
     return out;
 }
 
+void from_json(const nlohmann::json& json, TraceCall& cm) {
+    if (json.is_array()) {
+        cm.call = json.at(0);
+        cm.trace_config = json.at(1);
+    }
```
canepat, line 68: If input JSON is not an array, the `TraceCall` instance will be empty. Is it a valid value? If it is valid, shouldn't we explicitly accept an empty JSON value and reject other non-array values (e.g. a number followed by `null`)?

silkrpc/core/evm_trace.cpp
```diff
@@ -1133,7 +1165,7 @@ asio::awaitable<TraceCallResult> TraceCallExecutor<WorldState, VM>::execute(std:
     tracers.push_back(tracer);
 
     EVMExecutor<WorldState, VM> executor{io_context_, database_reader_, *chain_config_ptr, workers_, block_number};
-    for (auto idx = 0; idx < index; idx++) {
+    for (auto idx = 0; idx < transaction.transaction_index; idx++) {
```
canepat, line 1168: This `execute` method overload is used also by other APIs, so just to double-check: is it correct to use `transaction.transaction_index` instead of `index` here also for other APIs?

tests/integration/goerly/debug_traceTransaction/test_1.json
```diff
@@ -1,28 +0,0 @@
-[
```
canepat, line 1: May I ask why are we deleting this integration test?

tests/integration/goerly/trace_call/test_1.json
```diff
@@ -1,4773 +0,0 @@
-[
```
canepat, line 1: May I ask why are we deleting this integration test?

canepat (changes requested):

silkrpc/core/evm_trace.cpp
Sixtysixter: I did all checks as I know this method is used by several API. 
In addition it is better to use the `transaction_index `field to give `index` param additional meaning (ie, the one added by trace_callMany)

tests/integration/goerly/trace_call/test_1.json
Sixtysixter: As the standard (by @lupin012) is to use a double digit to enumerate test (from 00 to 99), the test_01.json is already present (maybe it was a refuse..)

tests/integration/goerly/debug_traceTransaction/test_1.json
Sixtysixter: See previous comment (there is some additional mistake to fix)

silkrpc/commands/trace_api.cpp
Sixtysixter: It was done following the way of all other methods that use `INFO` to show parameters...

silkrpc/core/evm_trace.cpp
Sixtysixter: You're right (it was not finished yet...). 
I decided to remove the if because it is better to leave `nlohmann::json` doing check and raise exception for us. 
By the way, we need a review in all case we parse the a wrong input JSON to get parameters as, in these cases, the response form `silkrcpdaemon` differ from `rcpdaemon` ones.

canepat (approved):

PR #357, opened 2022-10-20
Fix transaction compliant geth

lupin012:


tests/integration/run_tests.py
```diff
@@ -35,24 +36,36 @@
...
     """ determine target
     """
     if "engine_" in method:
         return "localhost:8550"
-    if silk:
+    if target_type == SILK:
         return "localhost:51515"
 
+    if target_type == INFURA:
+        return "https://goerli.infura.io/v3/d1c30a8114fd443aad1ed5d4b403e33d"
```
canepat, line 48: This URL must be passed as input argument and *NOT* hard-coded because:
- it is specific for Goerli network whilst this run script is not
- it contains personal account information

canepat (changes requested):

canepat (approved):

PR #379, opened 2022-11-19
Add JWT authentication to Engine API

lupin012:


silkrpc/http/request_handler.cpp
```diff
@@ -49,6 +77,59 @@ boost::asio::awaitable<void> RequestHandler::handle_request(const http::Request&
...
+                co_return;
+            }
+
+            try {
+               // Parse token
+               auto decoded_client_token = jwt::decode(client_token);
+
+               // Validate token
+               auto verifier = jwt::verify()
+                   .allow_algorithm(jwt::algorithm::hs256{convert_jwt_secret(jwt_secret_)});
```
canepat, line 115: I think `convert_jwt_secret` function is not needed: you can use `to_hex` and then convert it to `std::string` like this:
```
const auto jwt_secret_bytes = *from_hex(jwt_secret_); // must *not* fail (should check it)
auto verifier = jwt::verify()
    .allow_algorithm(jwt::algorithm::hs256{jwt_secret_bytes.cbegin(), jwt_secret_bytes.cend()});
```

Then, we can do better than this and avoid converting the JWT secret each time: just convert it once (checking the result) when generated/loaded in the `Daemon::run` and then just pass it as `const Bytes&` where needed.

So, basically all the classes have `const Bytes& jwt_secret_;` instead of `const std::string& jwt_secret_;`

canepat (changes requested):

silkrpc/http/server.hpp
```diff
@@ -68,6 +68,7 @@ class Server {
     boost::asio::ip::tcp::acceptor acceptor_;
 
     boost::asio::thread_pool& workers_;
+    std::string jwt_secret_;
```
canepat, line 71: In order to support both authenticated and unauthenticated `Server` instances, I suggest using `std::optional<std::string>` as type for `jwt_secret_` instead of using empty value: it expresses the programmer's intention much clearly

silkrpc/http/jwt_secret.cpp
```diff
@@ -0,0 +1,76 @@
+/*
```
canepat, line 1: Please remove `jwt_secret.cpp` since you have added `jwt.cpp`

silkrpc/http/jwt_secret.hpp
```diff
@@ -0,0 +1,39 @@
+/*
```
canepat, line 1: Please remove `jwt_secret.hpp` since you have added `jwt.hpp`

canepat (changes requested):

canepat (approved):
LTGM

PR #391, opened 2022-12-06
Response when id is missing in request

Sixtysixter:
Fixed response when the request doesn't contain the id field.


canepat (approved):
LGTM

PR #394, opened 2022-12-15
Support json batch

### LINKED ISSUE #98: Add support for JSON batch requests
Specification: [JSON Batch spec](https://www.jsonrpc.org/specification#batch)
Comparison: [Erigon RPCDaemon batch issue](https://github.com/ledgerwatch/erigon/issues/2089)

Implementation hints:
- `Request request_` -> `std::vector<Request> requests_` in `Connection`
- `Request& req` -> `std::vector<Request>& requests` in `RequestParser::parse`
- `RequestParser::consume` using i-th `Request` as long as `result` is not indeterminate and `begin != end` (`Parser::reset` before each iteration)
- `asio::co_spawn` to launch separate coroutine stacks in `Connection::do_read`, one for each `Request`

lupin012:
Closes #98


silkrpc/http/request_handler.cpp
```diff
@@ -47,18 +47,48 @@ boost::asio::awaitable<void> RequestHandler::handle_request(const http::Request&
...
-                reply.content = make_json_error(request_id, 403, error.value()).dump() + "\n";
-                reply.status = http::StatusType::unauthorized;
+
+        if (request_json.is_object()) {
+            if (!request_json.contains("id")) {
+                reply.content = "\n";
+                reply.status = http::StatusType::ok;
             } else {
-                co_await handle_request(request_json, reply);
+                auto request_id = request_json["id"].get<uint32_t>();
```
canepat, line 56: `request_id` can be made `const`

silkrpc/http/request_handler.cpp
```diff
@@ -47,18 +47,48 @@ boost::asio::awaitable<void> RequestHandler::handle_request(const http::Request&
...
+                const auto error = co_await is_request_authorized(request_id, request);
+                if (error.has_value()) {
+                    reply.content = make_json_error(request_id, 403, error.value()).dump() + "\n";
+                    reply.status = http::StatusType::unauthorized;
+                } else {
+                    co_await handle_request(request_json, reply, /* is_batch */ false);
```
canepat, line 62: The new parameter `is_batch` seems to be used just to append one newline at the end of the reply content when the request is not a batch one.
If I'm not missing something, you can obtain the same result in a simpler way:
- remove the `is_batch` parameter
- append one newline at the end of the reply content just after this `handle_request` call

silkrpc/http/request_handler.cpp
```diff
@@ -47,18 +47,48 @@ boost::asio::awaitable<void> RequestHandler::handle_request(const http::Request&
...
+                }
+            }
+        } else {
+            std::string full_json = "[";
+            bool first_element = true;
+            for (auto& el : request_json.items()) {
```
canepat, line 68: `el` -> `item`

silkrpc/http/request_handler.cpp
```diff
@@ -47,18 +47,48 @@ boost::asio::awaitable<void> RequestHandler::handle_request(const http::Request&
...
+                const auto element = el.value();
```
canepat, line 69: `element` -> `item_json`

silkrpc/http/request_handler.cpp
```diff
@@ -47,18 +47,48 @@ boost::asio::awaitable<void> RequestHandler::handle_request(const http::Request&
...
+                if (!element.contains("id")) {
+                    reply.content = "\n";
+                    reply.status = http::StatusType::ok;
+                } else {
+                    auto request_id = element["id"].get<uint32_t>();
```
canepat, line 74: `request_id` can be made `const`

silkrpc/http/request_handler.cpp
```diff
@@ -47,18 +47,48 @@ boost::asio::awaitable<void> RequestHandler::handle_request(const http::Request&
...
+                const auto error = co_await is_request_authorized(request_id, request);
+                if (error.has_value()) {
+                    reply.content = make_json_error(request_id, 403, error.value()).dump() + "\n";
+                    reply.status = http::StatusType::unauthorized;
+                } else {
+                    co_await handle_request(request_json, reply, /* is_batch */ false);
+                }
+            }
+        } else {
+            std::string full_json = "[";
```
canepat, line 66: `full_json` -> `batch_reply_content`

canepat (changes requested):

silkrpc/http/request_handler.cpp
lupin012: Yes it's better

canepat (approved):
LGTM

PR #396, opened 2022-12-17
Silkworm update

lupin012:
Updates to silkworm 15/12


silkrpc/concurrency/context_pool_test.cpp
```diff
@@ -57,7 +57,7 @@ TEST_CASE("Context", "[silkrpc][context_pool]") {
             Context context{create_channel, block_cache, state_cache, wait_mode};
             std::atomic_bool processed{false};
             auto* io_context = context.io_context();
-            io_context->post([&]() {
+            post(io_context->get_executor(), [&]() {
                 processed = true;
                 context.stop();
             });
```
canepat, line 63: I would change this as:
```
boost::asio::post(*io_context, [&]() {
    processed = true;
    context.stop();
});
```
because 1) it makes clearer which `post` function we're calling 2) it is more compact using the overload taking `boost::asio::context&` as first parameter.

silkrpc/concurrency/context_pool_test.cpp
```diff
@@ -69,7 +69,7 @@ TEST_CASE("Context", "[silkrpc][context_pool]") {
         SECTION(std::string("Context::stop wait_mode=") + std::to_string(static_cast<int>(wait_mode))) {
             Context context{create_channel, block_cache, state_cache, wait_mode};
             std::atomic_bool processed{false};
-            context.io_context()->post([&]() {
+            post(context.io_context()->get_executor(), [&]() {
                 processed = true;
             });
```
canepat, line 74: Same as above

canepat (changes requested):

silkrpc/concurrency/context_pool_test.cpp
lupin012: ok

silkrpc/concurrency/context_pool_test.cpp
lupin012: ok

canepat (approved):

PR #397, opened 2022-12-23
Get block number with latest

lupin012:


silkrpc/commands/debug_api.cpp
```diff
@@ -128,10 +128,10 @@ boost::asio::awaitable<void> DebugRpcApi::handle_debug_get_modified_accounts_by_
 
     try {
         ethdb::TransactionDatabase tx_database{*tx};
-        const auto start_block_number = co_await core::get_block_number(start_block_id, tx_database);
-        const auto end_block_number = co_await core::get_block_number(end_block_id, tx_database);
+        const auto start_block_number = co_await core::get_block_number(start_block_id, tx_database, false);
```
canepat, line 131: In order to increase readability and clarity, I suggest to use a comment with the name for the `latest_required` field at each `core::get_block_number` call site:
```
... = co_await core::get_block_number(start_block_id, tx_database, /*latest_required=*/false);
```

silkrpc/core/blocks.cpp
```diff
@@ -28,22 +28,32 @@ constexpr const char* kHeadBlockHash = "headBlockHash";
...
+        block_number.number = co_await get_latest_executed_block_number(reader);
+        block_number.is_latest_block = true;
     } else {
-        block_number = std::stol(block_id, 0, 0);
+        block_number.number = std::stol(block_id, 0, 0);
+        determine_is_latest_block = latest_is_required;
     }
-    SILKRPC_DEBUG << "get_block_number block_number: " << block_number << "\n";
+    if (determine_is_latest_block) {
+        block_number.is_latest_block = (block_number.number ==  co_await get_latest_executed_block_number(reader));
```
canepat, line 54: Probably, here you could use the new function `is_latest_block_number` suggested above

canepat (changes requested):

silkrpc/commands/debug_api.cpp
```diff
@@ -128,8 +128,8 @@ boost::asio::awaitable<void> DebugRpcApi::handle_debug_get_modified_accounts_by_
 
     try {
         ethdb::TransactionDatabase tx_database{*tx};
-        const auto start_block_number = co_await core::get_block_number(start_block_id, tx_database);
-        const auto end_block_number = co_await core::get_block_number(end_block_id, tx_database);
+        const auto [start_block_number, _] = co_await core::get_block_number(start_block_id, tx_database, /*latest_required=*/false);
+        const auto [end_block_number, ignore] = co_await core::get_block_number(end_block_id, tx_database, /*latest_required=*/false);
```
canepat, line 132: I've just now realized that we should also keep the previous `get_block_number` overload:
```
awaitable<uint64_t> get_block_number(const std::string& block_id, const DatabaseReader& reader);
```
and simply delegate to the new overload in its implementation:
```
awaitable<uint64_t> get_block_number(const std::string& block_id, const DatabaseReader& db_reader) {
    const auto [block_number, _] = co_await get_block_number(block_id, db_reader, /*latest_required=*/false);
    return block_number;
}
```
This would allow us to use the first overload version wherever we don't case about `is_latest` flag (i.e. where we currently pass `/*latest_required=*/false` and ignore the returned flag), making the code shorter and clearer at call site.

canepat (changes requested):

silkrpc/commands/debug_api.cpp
lupin012: ok

canepat (approved):
LGTM

PR #400, opened 2022-12-24
Add erigon_blockNumber API

### LINKED ISSUE #398: Add erigon_blockNumber

lupin012:
Closes #398


silkrpc/commands/erigon_api.cpp
```diff
@@ -324,4 +324,43 @@ boost::asio::awaitable<void> ErigonRpcApi::handle_erigon_watch_the_burn(const nl
     co_return;
 }
 
+
+// https://eth.wiki/json-rpc/API#erigon_blockNumber
+boost::asio::awaitable<void> ErigonRpcApi::handle_erigon_block_number(const nlohmann::json& request, nlohmann::json& reply) {
+    auto params = request["params"];
```
canepat, line 330: `auto params` -> `const auto params`

silkrpc/commands/erigon_api.cpp
```diff
@@ -324,4 +324,43 @@ boost::asio::awaitable<void> ErigonRpcApi::handle_erigon_watch_the_burn(const nl
...
+    if (params.size() == 0) {
+       block_id = core::kLatestExecutedBlockId;
+    } else if (params.size() > 1) {
+        auto error_msg = "invalid erigon_getBlockNumber params: " + params.dump();
+        SILKRPC_ERROR << error_msg << "\n";
+        reply = make_json_error(request["id"], 100, error_msg);
+        co_return;
+    } else {
+        block_id = params[0];
+    }
```
canepat, line 341: I suggest to switch the last two branches and adjust the condition accordingly:
```
    if (params.size() == 0) {
       block_id = core::kLatestExecutedBlockId;
    } else if (params.size() == 1) {
        block_id = params[0];
    } else {
        auto error_msg = "invalid erigon_getBlockNumber params: " + params.dump();
        SILKRPC_ERROR << error_msg << "\n";
        reply = make_json_error(request["id"], 100, error_msg);
        co_return;
    }
```
It seems more natural and also separates the valid parameter values (0, 1) from the invalid ones (all the other values)

silkrpc/commands/erigon_api.cpp
```diff
@@ -324,4 +324,43 @@ boost::asio::awaitable<void> ErigonRpcApi::handle_erigon_watch_the_burn(const nl
...
+    auto tx = co_await database_->begin();
+
+    try {
+        ethdb::TransactionDatabase tx_database{*tx};
+
+        const auto block_number{co_await core::get_block_number_by_tag(block_id, tx_database)};
+
+        std::ostringstream block_number_hex;
+        block_number_hex << std::hex << block_number;
+        reply = make_json_content(request["id"], "0x" + block_number_hex.str());
```
canepat, line 353: Shouldn't we use `QUANTITY` type format (i.e. `to_quantity` method) for the returned block number? It would be more coherent with other APIs

canepat (changes requested):

silkrpc/commands/erigon_api.cpp
lupin012: ok

silkrpc/commands/erigon_api.cpp
lupin012: ok

silkrpc/commands/erigon_api.cpp
lupin012: ok

canepat (approved):

PR #401, opened 2022-12-25
Fix build on VS2022 and test fixes

wezrule:
- `...` for ranges in case statements is not standard and available with the MSVC compiler
- `mimalloc` is enabled by default so updated readme to include that.
- `version_test.cpp` requires /bigobj specified to build it
- `erigon_api_test.cpp` Error:
`error C2248: 'silkrpc::commands::ErigonRpcApi::handle_erigon_get_block_by_timestamp': cannot access protected member declared in class 'silkrpc::commands::ErigonRpcApi'`
https://developercommunity.visualstudio.com/t/cannot-access-protected-member-declared-in-class/1666365

`erigon_api_test.cpp(154,10): error C2059: syntax error: 'constant'`
`error C2017: illegal escape sequence` 
Not sure why this is an issue, but moving it to a variable fixes it.

- `cxxabi.h` This is not available in MSVC, #ifdef'ed it out
- `SO_REUSEPORT` does not exist in Windows
- `cached_database_test.cpp` Lifetime issue with the string created for the table, could be issues elsewhere with this.
- `cursor.cpp` subscript access error

- Was a case insensitive issue with `Invalid`: `Receipt CBOR: missing entries: invalid argument`, seems to be platform dependent.
- "-std:c++20" compiler flag doesn't exist with MSVC


canepat:
Thank you very much for your contribution!

Currently none of us in the team uses Windows as development platform any more, so unfortunately full compatibility with it in the code base fall behind. IMO this is mainly due to the fact we lack one CI job for Windows.

@wezrule we would like to keep Windows compatibility fully functioning once restored by your PR, but we cannot make any guarantee without at least one CI job for Windows. So, if you are interested in using Silkrpc on Windows, probably you may consider to contribute one CI job for such platform?

I started PR #42 a long time ago to address this feature, but setting up a CI job for Windows without a local machine to test the build was too hard.

wezrule:
@canepat No problem!

I'm away for a few days now for the holidays but I can have a look after sure

canepat (approved):
LGTM

PR #403, opened 2022-12-29
Add erigon_cumulativeChainTraffic API

### LINKED ISSUE #404: Add erigon_cumulativeChainTraffic.

lupin012:
closes #404


silkrpc/commands/erigon_api.cpp
```diff
@@ -361,4 +359,42 @@ boost::asio::awaitable<void> ErigonRpcApi::handle_erigon_block_number(const nloh
...
+        chain_traffic.cumulative_gas_used  = co_await core::rawdb::read_cumulative_gas_used(tx_database, block_number);
+
+        reply = make_json_content(request["id"], chain_traffic);
+    } catch (const std::exception& e) {
+        SILKRPC_ERROR << "exception: " << e.what() << " processing request: " << request.dump() << "\n";
+        reply = make_json_content(request["id"], chain_traffic);
+    } catch (...) {
+        SILKRPC_ERROR << "unexpected exception processing request: " << request.dump() << "\n";
+        reply = make_json_error(request["id"], 100, "unexpected exception");
+    }
```
canepat, line 393: The JSON content returned in case of exception is kind of strange because it is a valid JSON result with an empty chain traffic data in case of any `std::exception`, whilst it is a JSON error in case of any other unorthodox exception.

Is this by any chance required to match Erigon behavior? If not, I would suggest a more consistent error handling.

lupin012: Yes to match erigon behaviour. In case of empty block the  response contains valid numbers not error.

canepat (approved):

PR #411, opened 2023-01-08
Add erigon_nodeInfo API implementation

### LINKED ISSUE #410: Add erigon_nodeInfo

lupin012:
closes #410


silkrpc/ethbackend/remote_backend.cpp
```diff
@@ -91,6 +93,33 @@ boost::asio::awaitable<uint64_t> RemoteBackEnd::net_peer_count() {
...
+        NodeInfo node_info;
+        const auto backend_node_info = reply.nodesinfo(i);
+        node_info.id = backend_node_info.id();
+        node_info.name = backend_node_info.name();
+        node_info.enode = backend_node_info.enode();
+        node_info.enr = backend_node_info.enr();
+        node_info.listenerAddr = backend_node_info.listeneraddr();
+        const auto protocols = backend_node_info.protocols();
+        auto protocols_json = nlohmann::json::parse(protocols, nullptr, /* allow_exceptions = */ false);
+        node_info.protocols = decode_protocols(protocols_json);
```
canepat, line 111: I suggest not to decode the protocol JSON fragment here if the same exact JSON object will be present in the `erigon_nodeInfo` JSON: you can just keep it as plain string and move the `nlohmann::json::parse` call to the place where JSON serialization of `erigon_nodeInfo` result happens.

silkrpc/ethbackend/remote_backend.cpp
```diff
@@ -393,4 +422,43 @@ std::string RemoteBackEnd::decode_status_message(const remote::EngineStatus& sta
     }
 }
 
+NodeInfoProtocols RemoteBackEnd::decode_protocols(const nlohmann::json& json) {
```
canepat, line 425: This method should be removed as it is unnecessary (see previous comment)

silkrpc/types/node_info.hpp
```diff
@@ -0,0 +1,81 @@
...
+    NodeInfoEthereum eth;
+};
+
+struct NodeInfo {
+    std::string       id;
+    std::string       name;
+    std::string       enode;
+    std::string       enr;
+    std::string       listenerAddr;
+    NodeInfoProtocols protocols;
```
canepat, line 75: As suggested above, I would make `protocols` a plain string here because parsing is unnecessary. This will also allow to remove `NodeInfoProtocols`, `NodeInfoEthereum`, `NodeInfoConfig` and `NodeInfoClique`.

canepat (changes requested):

silkrpc/ethbackend/remote_backend.cpp
lupin012: ok

silkrpc/types/node_info.hpp
lupin012: ok

silkrpc/ethbackend/remote_backend.cpp
lupin012: ok

canepat (approved):

PR #413, opened 2023-01-13
Fix set beneficiary on evm

lupin012:


silkrpc/consensus/ethash.cpp
```diff
@@ -27,7 +27,7 @@ BlockReward compute_reward(const ChainConfig& config, const silkworm::Block& blo
         throw std::runtime_error("Invalid chain config");
     }
     const auto chain_config = cc_optional.value();
-    const auto revision = chain_config.revision(block.header.number);
+    const auto revision = chain_config.revision(block.header.number, 0);
```
canepat, line 30: Please add a comment to make the second actual parameter more readable:
```
const auto revision = chain_config.revision(block.header.number, /*block_time=*/0);
```

silkrpc/core/evm_executor.hpp
```diff
@@ -81,6 +85,7 @@ class EVMExecutor {
     boost::asio::thread_pool& workers_;
     state::RemoteState& remote_state_;
     WorldState state_;
+    std::unique_ptr<silkworm::consensus::IEngine> _consensus_engine;
```
canepat, line 88: According to our name conventions, the leading underscore should be a trailing one (see also other member variables here): `consensus_engine_`

:smile:

canepat (changes requested):

silkrpc/core/evm_executor.hpp
```diff
@@ -62,7 +63,10 @@ class EVMExecutor {
         boost::asio::thread_pool& workers,
         uint64_t block_number,
         state::RemoteState& remote_state)
-        : io_context_(io_context), db_reader_(db_reader), config_(config), workers_{workers}, remote_state_{remote_state}, state_{remote_state_} {}
+        : io_context_(io_context), db_reader_(db_reader), config_(config), workers_{workers}, remote_state_{remote_state}, state_{remote_state_} {
+             _consensus_engine = silkworm::consensus::engine_factory(config);
+             assert(_consensus_engine != NULL);
```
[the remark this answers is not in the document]
lupin012: ok

silkrpc/consensus/ethash.cpp
lupin012: ok

silkrpc/core/evm_executor.hpp
lupin012: typo.

canepat (approved):

PR #414, opened 2023-01-22
Fix git branch info in perf test runner

lupin012:


canepat (approved):
I suggest to try and use [GitPython](https://github.com/gitpython-developers/GitPython) instead of shell commands in a future PR to amend this

PR #415, opened 2023-01-24
Add ots_getApiLevel and ots_hasCode API implementation

dario1995:


silkrpc/commands/ots_api.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+namespace silkrpc::commands {
+
+boost::asio::awaitable<void> OtsRpcApi::handle_ots_get_api_level(const nlohmann::json& request, nlohmann::json& reply) {
+    reply = make_json_content(request["id"], 8);
+    co_return;
+}
+
+boost::asio::awaitable<void> OtsRpcApi::handle_ots_has_code(const nlohmann::json& request, nlohmann::json& reply) {
+    auto params = request["params"];
```
canepat, line 34: `params` can be `const` here

silkrpc/commands/ots_api.cpp
```diff
@@ -0,0 +1,76 @@
...
+    if (params.size() != 2) {
+        auto error_msg = "invalid ots_hasCode params: " + params.dump();
```
canepat, line 36: `error_msg` can be `const`

silkrpc/commands/ots_api.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+#include "silkrpc/core/blocks.hpp"
+#include "silkrpc/core/state_reader.hpp"
+#include "silkrpc/ethdb/kv/cached_database.hpp"
+#include "silkrpc/ethdb/transaction_database.hpp"
+
+namespace silkrpc::commands {
+
+boost::asio::awaitable<void> OtsRpcApi::handle_ots_get_api_level(const nlohmann::json& request, nlohmann::json& reply) {
+    reply = make_json_content(request["id"], 8);
```
canepat, line 29: Please add a constant here instead of hard-coded value, something like:
```
constexpr int kCurrentApiLevel{8};
```
You can put the constant inside the namespace before the `handle_ots_get_api_level` method definition.

canepat (changes requested):

silkrpc/commands/ots_api.cpp
dario1995: Ok

canepat (approved):

PR #417, opened 2023-01-25
Update API document

canepat:


docs/API.md
```diff
@@ -72,22 +75,27 @@ The following table shows the current [JSON RPC API](https://eth.wiki/json-rpc/A
...
 | engine_exchangeTransitionConfigurationV1   | Yes          |                                            |
 |                                            |              |                                            |
 | debug_accountRange                         | Yes          |                                            |
+| debug_accountAt                            | -            | not yet implemented                        |
 | debug_getModifiedAccountsByHash            | Yes          |                                            |
 | debug_getModifiedAccountsByNumber          | Yes          |                                            |
 | debug_storageRangeAt                       | Yes          |                                            |
 | debug_traceBlockByHash                     | Yes          |                                            |
 | debug_traceBlockByNumber                   | Yes          |                                            |
-| debug_traceTransaction                     | Yes          |                                            |
```
Sixtysixter, line 89: Saying 'uses JSON streaming' is a short but not fully correct way to describe the actual implementation, that's based on http chunked transfer encoding (see [https://en.wikipedia.org/wiki/Chunked_transfer_encoding]). 
May be we could state:

> uses chunked transfer encoding in http reply

 ?

canepat: You may have a point here but I would prefer to use `JSON streaming` because it matches the terminology used in Erigon

Sixtysixter: OK

Sixtysixter (approved):

